PR #124, opened 2020-02-21
report eastwood detail urls

thumbnail:
## Brief

the key is a string. conforming it to a keyword. 

Now it'l show up in our reports :)

## QA plan

green tests

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [ ] Modular design
  * [x] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [x] I have checked out this branch and reviewed it locally, running it
* [x] I have reviewed the PR
  * [x] Correctness
  * [ ] Modular design
  * [x] Test coverage
  * [x] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -36,15 +36,16 @@
                    (= "*" ;; False positives for dynamic vars https://git.io/fhQTx
                       (-> v str first)
                       (-> v str last))))
-         (map (fn [{:keys [uri-or-file-name linter] :as m}]
+         (map (fn [{:keys [uri-or-file-name linter] :strs [warning-details-url] :as m}]
                 (assoc m
-                       :level    :warning
-                       :source   (keyword "eastwood" (name linter))
-                       :filename (if (string? uri-or-file-name)
-                                   uri-or-file-name
-                                   (str/replace (-> ^File uri-or-file-name .getPath)
-                                                root-dir
-                                                "")))))
+                       :level               :warning
+                       :source              (keyword "eastwood" (name linter))
+                       :warning-details-url warning-details-url
+                       :filename            (if (string? uri-or-file-name)
+                                              uri-or-file-name
+                                              (str/replace (-> ^File uri-or-file-name .getPath)
+                                                           root-dir
+                                                           "")))))
          (concat (impl/warnings->reports output)))))
 
 (defn new [{:keys [eastwood-options]
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -6,7 +6,7 @@
    [formatting-stack.linters.eastwood.impl :as impl]
    [formatting-stack.protocols.linter :as linter]
    [formatting-stack.util :refer [ns-name-from-filename]]
-   [medley.core :refer [deep-merge]]
+   [medley.core :refer [assoc-some deep-merge]]
    [nedap.utils.modular.api :refer [implement]])
   (:import
    (java.io File)))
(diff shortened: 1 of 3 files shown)


src/formatting_stack/linters/eastwood.clj
```diff
@@ -36,15 +36,16 @@
...
-                       :level    :warning
-                       :source   (keyword "eastwood" (name linter))
-                       :filename (if (string? uri-or-file-name)
-                                   uri-or-file-name
-                                   (str/replace (-> ^File uri-or-file-name .getPath)
-                                                root-dir
-                                                "")))))
+                       :level               :warning
+                       :source              (keyword "eastwood" (name linter))
+                       :warning-details-url warning-details-url
```
vemv: I'd assoc it only if `(seq warning-details-url)`

(we cannot presume an external library will always provide this key for each linters, paricularly as the lib evolves)

vemv: :warning-details-url is missing in `formatting-stack.protocols.spec` right?

test/functional/formatting_stack/linters/eastwood.clj
```diff
@@ -25,10 +25,11 @@
...
          :column   25
          :filename "test-resources/eastwood_warning.clj"}
-        {:source   :eastwood/def-in-def
-         :line     3
-         :column   13
-         :filename "test-resources/eastwood_warning.clj"}
+        {:source              :eastwood/def-in-def
+         :line                3
+         :column              13
+         :warning-details-url "https://github.com/jonase/eastwood#def-in-def"
```
vemv: I find it interesting that `match?` allows the other test cases within this `are` to omit `:warning-details-url` and still pass.

I get why that's like that, but it also leads to:

* test case heterogeneity
  * having different shapes of data is confusing - it makes me wonder "why the differences? do they have some intent relevant to the thing being tested?"
* currently only a fraction of the tests in this `are` exercise the `:warning-details-url` aspect
  * that literally means we get less coverage
  * it also means that fewer tests act as documentation, which is an important role
* fragility
  * I can delete random key-value entries in these maps and have the tests still pass
    * The same cannot be said of most testing techniques/frameworks I know

As a tentative conclusion, `match?` is too DRY, up to the point of self-reference.

IMO a good test compares `x` against `y`. If there's only an `x` and no `y`, then `x` becomes a self-referential test (input == expectation).

Of course, https://github.com/nedap/formatting-stack/pull/124 is not critical functionality, but as you know I've always been a bit skeptical about `match?` as a general-purpose tool. Maybe we can begin limiting its usage elsewhere.

thumbnail: > I find it interesting that `match?` allows the other test cases within this `are` to omit `:warning-details-url` and still pass.

The effect you're describing is only one of the matchers (`matchers/embeds`, default for maps). This behavior is fully configurable. 

The default is set to embeds because maps are open to extension exact matches,. Breakage tend to be cumbersome and overwhelming when every `are`-case fails (e.g. because an extra key is added)

> having different shapes of data is confusing 

I'd say it might be unfamiliar. I've come to expect these shapes for the big-are clauses which leverage `match?`

> currently only a fraction of the tests in this are exercise the `:warning-details-url` aspec
    -  that literally means we get less coverage

It means we get less _duplicated_ coverage, which leads to more specific failures. one `are`-case failing helps pin-pointing the issue.

>  I can delete random key-value entries in these maps and have the tests still pass

From the expected-values, yes. the actual values, no.

> The same cannot be said of most testing techniques/frameworks I know

It's rather common in Ruby's rspec ([`have-attributes-matcher`](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/have-attributes-matcher), [`contains_exactly`](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/contain-exactly-matcher) (similar to `m/equals`), and [`include`](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/include-matcher)), Java's hamcrest (`IsIterableContaining`, `anyOf`, `arrayContaining`, see [matchers](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html)), python as a hamcrest impl as well.

PR #137, opened 2020-02-27
Gracefully handle Git deletions again

vemv:
## Brief

`impl/absolutize`, particularly with its `.exists` assertion,  had broken our ability to gracefully handle deleted files (staged or not).

This PR fixes that, adding an extensive integration test.

## QA plan

Test all cases in a real project (suggestion: a monorepo):

* Deleted file (staged for deletion or not)
* plain `core/format!`, `format-and-lint-project!`, `format-and-lint-branch!`

That's 6 scenarios - please let's test them all (I have in formatting-stack itself).

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Modular design
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
    * `defn all-files` peforms 3 git ops but they're fairly cheap (5-15ms each)
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -10,7 +10,7 @@
 
   A strategy may not return nil."
   (:require
-   [clojure.string :as str]
+   [clojure.string :as string]
    [clojure.tools.namespace.repl :refer [refresh-dirs]]
    [formatting-stack.protocols.spec :as protocols.spec]
    [formatting-stack.strategies.impl :as impl]
@@ -28,10 +28,18 @@
   (let [tracked (->> (impl/file-entries git-command "ls-files")
                      (impl/absolutize git-command))
         untracked (->> (impl/file-entries git-command "ls-files" "--others" "--exclude-standard")
-                       (impl/absolutize git-command))]
+                       (impl/absolutize git-command))
+
+        deleted (binding [impl/*skip-existing-files-check?* true]
+                  (->> (impl/file-entries git-command "status" "--porcelain")
+                       (filter impl/deleted-file?)
+                       (map impl/remove-deletion-markers)
+                       (impl/absolutize git-command)))]
     (->> files
          (into tracked)
          (into untracked)
+         (remove (set deleted))
+         (impl/absolutize git-command)
          (impl/extract-clj-files))))
 
 (speced/defn git-completely-staged
@@ -40,10 +48,11 @@
       :or   {impl (impl/file-entries git-command "status" "--porcelain")}}]
   (->> impl
        (filter #(re-find impl/git-completely-staged-regex %))
-       (map #(str/replace-first % impl/git-completely-staged-regex ""))
+       (remove impl/deleted-file?)
+       (map #(string/replace-first % impl/git-completely-staged-regex ""))
        (map (fn [s]
               ;; for renames:
-              (-> s (str/split #" -> ") last)))
+              (-> s (string/split #" -> ") last)))
        (impl/absolutize git-command)
        (impl/extract-clj-files)
        (into files)))
@@ -53,8 +62,9 @@
   [& {:keys [^::protocols.spec/filenames files, impl]
       :or   {impl (impl/file-entries git-command "status" "--porcelain")}}]
   (->> impl
+       (remove impl/deleted-file?)
        (filter #(re-find impl/git-not-completely-staged-regex %))
-       (map #(str/replace-first % impl/git-not-completely-staged-regex ""))
+       (map #(string/replace-first % impl/git-not-completely-staged-regex ""))
        (impl/absolutize git-command)
        (impl/extract-clj-files)
        (into files)))
@@ -64,9 +74,10 @@
   The diff is compared against the `:target-branch` option."
   [& {:keys [target-branch impl files blacklist]
       :or   {target-branch "master"
-             impl          (impl/file-entries git-command "diff" "--name-only" target-branch)
+             impl          (impl/file-entries git-command "diff" "--name-status" target-branch)
              blacklist     (git-not-completely-staged :files [])}}]
   (->> impl
+       (remove impl/deleted-file?)
        (impl/absolutize git-command)
        (remove (set blacklist))
        (impl/extract-clj-files)
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -36,10 +36,10 @@
                        (map git-status/remove-deletion-markers)
                        (impl/absolutize git-command)
                        (set))
-          tracked (->> (impl/file-entries git-command "ls-files")
+          tracked (->> (impl/file-entries git-command "ls-files" "--full-name")
                        (impl/absolutize git-command)
                        (remove deleted))
-          untracked (->> (impl/file-entries git-command "ls-files" "--others" "--exclude-standard")
+          untracked (->> (impl/file-entries git-command "ls-files" "--full-name" "--others" "--exclude-standard")
                          (impl/absolutize git-command)
                          (remove deleted))]
       ;; Second `binding`, to ensure correct results
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -1,6 +1,8 @@
 (ns formatting-stack.strategies
   "Strategies are concerned with generating a seq of filenames to process (format, lint, or compile).
 
+  Said filenames must follow `::protocols.spec/filenames`, and represent an existing file.
+
   They are configured to run in a determinate order.
 
   In practice, a strategy is function that receives a seq of filenames, and returns another:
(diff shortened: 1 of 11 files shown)


src/formatting_stack/strategies.clj
```diff
@@ -25,25 +28,37 @@
...
+                       (set))
+          tracked (->> (impl/file-entries git-command "ls-files")
+                       (impl/absolutize git-command)
+                       (remove deleted))
+          untracked (->> (impl/file-entries git-command "ls-files" "--others" "--exclude-standard")
+                         (impl/absolutize git-command)
+                         (remove deleted))]
+      ;; Second `binding`, to ensure correct results
+      (binding [impl/*skip-existing-files-check?* false]
+        (speced/let [^::impl/existing-files corpus (into tracked untracked)]
```
thumbnail: you could remove some duplication by doing this:

```suggestion
          tracked   (impl/file-entries git-command "ls-files")
          untracked (impl/file-entries git-command "ls-files" "--others" "--exclude-standard")]
      ;; Second `binding`, to ensure correct results
      (binding [impl/*skip-existing-files-check?* false]
        (speced/let [^::impl/existing-files corpus (->> (into tracked untracked)
                                                        (impl/absolutize git-command)
                                                        (remove deleted))]
```

src/formatting_stack/strategies.clj
```diff
@@ -64,9 +80,12 @@
   The diff is compared against the `:target-branch` option."
   [& {:keys [target-branch impl files blacklist]
       :or   {target-branch "master"
-             impl          (impl/file-entries git-command "diff" "--name-only" target-branch)
+             impl          (impl/file-entries git-command "diff" "--name-status" target-branch)
              blacklist     (git-not-completely-staged :files [])}}]
   (->> impl
+       (remove (fn [s]
+                 (-> s (string/starts-with? "D\t"))))
```
thumbnail: We should use `git-status/deleted-file?` here

src/formatting_stack/strategies.clj
```diff
@@ -64,9 +80,12 @@
...
+       (map #(string/replace-first % #"(A|C|D|M|R|T|U|X|B)\t" "")) ;; See: `git diff --help`, `diff-filter` section
```
thumbnail: What about removing the first _word_ from the output? it seems less brittle

test/integration/formatting_stack/strategies.clj
```diff
@@ -0,0 +1,153 @@
...
+(def ^File deletable-file (File. "test-resources/deletable.clj"))
+
+(def ^String createable-filename "test-resources/createable.clj")
+
+(def creatable-contents (pr-str '(ns foo)))
+
+(assert (-> deletable-file .exists))
+
+(assert (not (-> createable-filename File. .exists)))
+
```
thumbnail: These should be inside of the test. The files don't have to exist/not-exists when _loading_ this ns, they should when i run the tests

test/integration/formatting_stack/strategies.clj
```diff
@@ -0,0 +1,153 @@
...
+        (sh "git" "checkout" (str deletable-file))
+        (assert (-> deletable-file .exists))))))
+
+(deftest git-diff-against-default-branch
+
+  (assert-pristine-git-status!)
+
+  (testing "It runs without errors, exercising its specs, even in face of deleted (but not staged) files"
+    (let [;; f-stack's first ever commit. This ensures a large, diverse corpus:
+          first-commit "ebb0ca8"]
```
thumbnail: could be [`git rev-list --max-parents=0 HEAD`](https://stackoverflow.com/questions/5188914/how-to-show-first-commit-by-git-log). to keep the actual commit hash out of the source

thumbnail: The let can be one level higher, (or a delayed def with the command above) preventing duplication on L77

test/integration/formatting_stack/strategies.clj
```diff
@@ -0,0 +1,153 @@
...
+      (testing "Added files don't show up"
+        (try
+          (sh "git" "add" "-A")
+          (is (= []
+                 (sut/git-not-completely-staged :files [])))
+          (finally
+            (sh "git" "reset" "."))))
+
+      (finally
+        (-> createable-filename File. .delete))))
```
thumbnail: exceptions will be caught inside of the `is`, right? does this work how the way its intended?

src/formatting_stack/strategies.clj
vemv: > you could remove some duplication by doing this:

Will try

src/formatting_stack/strategies.clj
vemv: > We should use git-status/deleted-file? here

Nope, note `git-status`' ns docstring

src/formatting_stack/strategies.clj
vemv: > What about removing the first word from the output? i

Dunno, removing arbitrary stuff can also be brittle.

unit-testing the `replace-first` can be a middle ground

test/integration/formatting_stack/strategies.clj
vemv: > These should be inside of the test.

Note that 2 deftests depend on this assertion. Having 2 deftests fail seems less clear, concise and immediate than having `(refresh)` fail as soon as an issue arises.

test/integration/formatting_stack/strategies.clj
vemv: > git rev-list --max-parents=0 HEAD.

Will try

test/integration/formatting_stack/strategies.clj
vemv: > exceptions will be caught inside of the is, right?

`finally` runs 100% unconditionally, whether an exception was thrown or not.

test/integration/formatting_stack/strategies.clj
thumbnail: Except that now i can refresh, delete the file and run the test. yielding unclear failures as well.

src/formatting_stack/strategies.clj
thumbnail: ouch, okay :P

test/integration/formatting_stack/strategies.clj
vemv: > Except that now i can refresh, delete the file and run the test.

Running tests without a prior `(refresh)` seems just wrong. Our dev.clj file typically suggests to `(refresh)` prior to running any test

src/formatting_stack/strategies.clj
vemv: Cannot be done: the two `(impl/absolutize git-command)` calls must be called within the former binding.

Thankfully, `integration.formatting-stack.strategies` fails if we applied this patch

vemv:
Ready again!

vemv:
👀!

Did you get exceptions? Supposedly there's a spec that would assert `impl/absolutize` output would correspond to an existing file

thumbnail:
> Supposedly there's a spec that would assert impl/absolutize output would correspond to an existing file

It's probably happening when `*skip-existing-files-check?*` is bound to `true`

vemv:
> Did you get exceptions?

(For clarity: at least the spec checking part was working, w/ nice Expound reports)

thumbnail:
> Did you get exceptions?

Only the `ExceptionInfo`-exceptions with the validation errors.

<details>

```
-- Relevant specs -------

:formatting-stack.strategies.impl/existing-files:
  (clojure.spec.alpha/coll-of
   (nedap.speced.def/fn
    [s]
    (if
     formatting-stack.strategies.impl/*skip-existing-files-check?*
     true
     (clojure.core/-> s File. .exists))))

-- Spec failed --------------------

  (...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   "<SNIP>"
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...
   ...)

should satisfy

  (speced/fn
   [s]
   (if *skip-existing-files-check?* true (-> s File. .exists)))

-- Relevant specs -------

:formatting-stack.strategies.impl/existing-files:
  (clojure.spec.alpha/coll-of
   (nedap.speced.def/fn
    [s]
    (if
     formatting-stack.strategies.impl/*skip-existing-files-check?*
     true
     (clojure.core/-> s File. .exists))))

-------------------------
Detected 27 errors

Execution error (ExceptionInfo) at formatting-stack.strategies/all-files$fn (strategies.clj:47).
Validation failed
```

</details>

vemv:
Some notes:

* `git status porcelain` always emits paths relative to the git root, not to the cwd
* `git diff --name-status` always emits paths relative to the git root, not to the cwd
* `git ls-files`, at least by default emits paths relative to the cwd

The latter, beng different, is a source of issues.

vemv:
Kindly QA again. I'll be also, over a non-monorepo

thumbnail:
staged/unstaged deleted files within the same module work as expected. However `strategies/all-files` only returns the files present in `(System/getProperty "user.dir")` (e.g. the module itself). Not the files outside of the module.

vemv:
I thought about that.

It seems bit of a poor default to return files outside of the monorepo. Most folks will _not_ want files outside of the current project to be returned.

Else they will be formatting/linting files outside of what Clojure can `require`, so bugs would arise.

That's a non-issue for us `:monorepo` profile, or people doing similar stuff (https://clubhouse.io/blog/monolith-meet-mono-repo/). But it seems unsafe to assume that everyone will have that set up.

(Even among us :monorepo is optional)

thumbnail:
> Else they will be formatting/linting files outside of what Clojure can require, so bugs would arise.

This is not necessarily true, the files might be on the classpath, but not in the current project. It might be specific to our monorepo setup, but running `project-formatter/format-and-lint-project!` for each project could make sense 🤔

vemv:
> This is not necessarily true, the files might be on the classpath, but not in the current project.

Yes, but we cannot know with all certainty.

(Monorepo != "all projects can reach each other's set of namespaces")

Created https://github.com/nedap/formatting-stack/issues/139 + https://github.com/nedap/formatting-stack/issues/140 btw, which can be attended promptly but would not consider a blocker (as f-s is somewhat broken rn)

PR #145, opened 2020-03-05
Fix linting renamed files

thumbnail:
## Brief

Renaming files would yield existing-files errors when using branch-formatter

Using `git diff --diff-filtes` allows us to filter files with a specific status, in our case we only want Added, Copied, Modified and Renamed files. (excludes files which are unmerged, had a typechange, are unknown, have/had broken pairing, and are deleted)

## QA plan

 1. checkout a project with f-s
 1. rename & stage file
 1. run branch-formatter
 1. normal linting as expected

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Modular design
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes


vemv:
Looking great! Thanks for linking to https://git-scm.com/docs/git-status#_short_format as well.

Can you check out `git diff --help` and verify that that info is consistent?

(status !== diff)

It might also reveal some unforeseen corner case.

test/unit/formatting_stack/strategies/impl/git_diff.clj
```diff
@@ -30,6 +30,17 @@
     "B\tfoo.clj" "foo.clj"
     "foo.cljB\t" "foo.cljB\t"
 
+    ;; merge conflicts are replaced
+    "DU\t"        ""
+    "DU\tfoo.clj" "foo.clj"
+    "foo.cljDU\t" "foo.cljDU\t"
+
+    ;; `R` is replaced including optional previous path
+    "R001\tfoo.clj"          "foo.clj"
```
vemv: What's the output for `"R001R001\tfoo.clj"`?

I'd be wary in the case `R001R001` was trimmed. b/c it's not a real git input, and we shouldn't create confusing outputs (which would hinder future debugging, etc)

src/formatting_stack/strategies/impl/git_diff.clj
```diff
@@ -11,4 +11,4 @@
 
 ;; See: `git diff --help`, `diff-filter` section
 (speced/defn remove-markers [^string? s]
-  (-> s (string/replace-first #"^(A|C|D|M|R|T|U|X|B)\t" "")))
+  (-> s (string/replace-first #"^(A|C|D|M|R|T|U|X|B|\?|\!).*\t" "")))
```
vemv: I guess the changes also target these:

image

Include them in unit tests pls

thumbnail:
> Can you check out git diff --help and verify that that info is consistent?

`git diff --name-status` refers to `--diff-filter` docs; which only list the individual options (not the combinations). I used https://stackoverflow.com/a/54829033 as a basis for this change.

src/formatting_stack/strategies/impl/git_diff.clj
thumbnail: Wasn't added because of the comment `;; `B` is replaced (and so on for other letters - we're not testing a whole regex range)`. will add anyway though.

test/unit/formatting_stack/strategies/impl/git_diff.clj
thumbnail: > I'd be wary in the case R001R001 was trimmed. 

That's the case because everthing between `R` and `\t` is stripped. this is intended as it contains a random int and it's previous filepath.

The whole regex is a bit brittle as is, but i cant really find a more proper git-tool to find the information we're looking for anyway.

src/formatting_stack/strategies/impl/git_diff.clj
vemv: I see! At the same time, the requirements have changed hence the extra caution.

test/unit/formatting_stack/strategies/impl/git_diff.clj
vemv: Mmmm, I cannot have enough confidence in the proposed changes atm. A single `\t` shouldn't cause f-s to eat arbitrary inputs.

Today's bug shows that there are always "unknown unknowns", in which deterministic/delimited behavior helps.

> The whole regex is a bit brittle as is

No particular need for the thing to be a regex (or at least a _single_ regex). This is also fine: `(-> s (string/replace regex-or-str) (string/replace regex-or-str) (string/replace regex-or-str))` (point being: N calls can be acceptable vs. a contrived regex)

Needless to say, existing tests increase our confidence in swapping impls :)

thumbnail: >  A single \t shouldn't cause f-s to eat arbitrary inputs.

agreed, it works because coincidence rather than design.

Having multiple regexes without a clear doc on all possible outputs also seems problematic. Am working on a suggestion using `diff-filter` which should be more stable.

vemv:
If I removed the `--diff-filter=ACMR` bit, would the suite still pass?

thumbnail:
> If I removed the --diff-filter=ACMR bit, would the suite still pass?

I'd assume so. We have no tests covering any of the git commands

vemv:
> We have no tests covering any of the git commands

?

We have integration + unit ones.

The point of my question is - bugfixes should be accompanied by tests, avoiding having parts that we can unawarerly undo in a future.

thumbnail:
>`We have integration ones

Must've missed those, `integration.formatting-stack.strategies/git-not-completely-staged` fails if the `diff-filter`-param is omitted, added a test-case for the rename-example.

src/formatting_stack/strategies.clj
```diff
@@ -81,11 +80,9 @@
   The diff is compared against the `:target-branch` option."
   [& {:keys [target-branch impl files blacklist]
       :or   {target-branch "master"
-             impl          (impl/file-entries git-command "diff" "--name-status" target-branch)
+             impl          (impl/file-entries git-command "diff" "--name-only" "--diff-filter=ACMR" target-branch)
```
vemv: If I change ACMR -> ACMRTUXB the suite also passes, which might indicate some missing coverage?

More generally, why did you pick ACMR? I'd annotate it with a comment, similarly to https://github.com/nedap/formatting-stack/blob/5c6e1f11c6ec80475063df52e1aceaded8375aee/src/formatting_stack/strategies/impl/git_diff.clj#L12

i.e. opaque bits generally call for an explanation.

thumbnail: > If I change ACMR -> ACMRTUXB the suite also passes, which might indicate some missing coverage?

There's no tests for type-changed, unmerged, unknown and broken (pair) files. According to https://stackoverflow.com/a/6068691 there's no way to create files in a Unknown state.  For both broken, unmerged and type-changed files I could create a test. but it's rather cumbersome to get the repo in the exact state we need for the test. It also feels like i'm testing `git diff --diff-filter` instead of f-s.

---

I'll add a comment with the description of the PR regarding the choice

thumbnail:
I'm not sure how we can proceed to get this bugfix live.

vemv:
Will QA + give one last 👀later today

Btw, PR's impl changed; did you QA the newer impl (multiple cases, multiple types of repos)?

thumbnail:
> Will QA + give one last 👀later today

👍 

> Btw, PR's impl changed; did you QA the newer impl (multiple cases, multiple types of repos)?

I've successfully used `v4.1.1-alpha1` in a project with renamed and deleted files. Will try in a monorepo later today.

vemv:
I have verified that `git diff --name-only --diff-filter=ACMR master` always emits the same output regardless of the cwd (which is desirable).

Issue #134, opened 2020-02-24
duplicated defaults


thumbnail:
## Brief

defaults are duped over https://github.com/nedap/formatting-stack/blob/master/src/formatting_stack/defaults.clj, https://github.com/nedap/formatting-stack/blob/master/src/formatting_stack/branch_formatter.clj and https://github.com/nedap/formatting-stack/blob/master/src/formatting_stack/project_formatter.clj with minor (but existing) differences.

## Expected behavior

All defaults live in https://github.com/nedap/formatting-stack/blob/master/src/formatting_stack/defaults.clj

## Actual behavior

defaults are duped

## Suspected cause

differences in configuration between core/project-formatter/branch-formatter

## Environment info

<= v4.0.0

## Additional links

Thanks to #130 we can now apply the configuration differences needed in project-formatter/branch-formatter to the defaults.

vemv:
Maybe the protocols could some new methods:

`(defprotocol Formatter (default-strategies [_] "The default strategies that this forrmatter tends to need"))`

That way the duplication would be killed. One still should be able to override the default behavior.

thumbnail:
> Maybe the protocols could some new methods:

Interesting approach. Some linters still do some filtering themselves (remove edn files, unrequireable files, blacklisting known-bad files, etc.). Maybe it's more suited as part of #37 


> That way the duplication would be killed.

The simplest approach would be to remove the defaults in project-formatter and branch-formatter, and depend on the defaults in `formatting-stack.defaults`, overriding the defaults where necessary.

vemv:
Traditionally I've seen this as a non-issue since "it's just data".

However I'm starting to see the value/need for this, as I've tried to customize heavly one real project. Some parts could be customized as the [examples](https://github.com/nedap/formatting-stack/blob/14b910091e5c0889fa4ffa6ef2b1789508ffbbe9/test-resources/customization_example.clj) advertise; others needed more knowledge of the f-s impl.

At the same time it's quite crucial to keep f-s simple/easy (for us and for any casual user). I'd be wary of some DSL, hooks, DAG libs, etc

e.g. an API resembling a `deep-merge` would seem simple:

```clj
{:in-background? false
 :overrides {:linters {::kondo/id {:kondo-clj-options {:linters {:cond-else {:level :warning}}}}}
             :formatters {::newlines/id nil ;; remove it
                          ::cljfmt/id {:third-party-indent-specs {`a :defn}}}}}
```

[...]

PR #152, opened 2020-04-01
Introduce `:overrides` option

vemv:
## Brief

Implements https://github.com/nedap/formatting-stack/issues/134#issuecomment-590701727

Starting from this PR, `:overrides` becomes the primary API for customizing things: as a consumer, one should:

* Use one of: `git-status-formatter`, `branch-formatter`, `project-formatter`
* Pass `:overrides` to those if wanting to tweak something.

...and should not:

* Use `core/format!` directly
* Use the "factory" stuff introduced in this PR
 
## TODO 

* Address XXXs
* At this point, `git-status-formatter`, `branch-formatter`, `project-formatter` all deserve some test coverage.
* Explore how much can we DRY `git-status-formatter`, `branch-formatter`, `project-formatter`
  * A nice 'overrides' implementation would allow absolute DRY.
  * Might be less prioritary for now
* Create `impl` sub-nses for `git-status-formatter`, `branch-formatter`, `project-formatter` 
  * The intended API is `format-and-lint!` /  `lint!`
   * Things like `speced/defn default-formatters` are certainly not public
* Deprecate old APIs (kwargs) via `check!`
* Review doc (README)
* Review examples (test-resources)

## QA plan

* `git-status-formatter`, `branch-formatter`, `project-formatter` all keep working
* The examples work as advertised, customizing things as its code shows

## Author checklist

* [ ] The build passes
  * [ ] Correctness
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes


test/unit/formatting_stack/impl/overrides.clj
```diff
@@ -0,0 +1,97 @@
...
+
+(deftest apply-overrides
+  (let [sample-members [{:id :a/id :age 42 :bar {:baz 42}}]
+        reified-member (reify)]
+    (are [desc members overrides expected] (testing (pr-str [desc members overrides])
+                                             (is (= expected
+                                                    (try
+                                                      (sut/apply-overrides members overrides)
+                                                      (catch AssertionError _
+                                                        ::EXCEPTION))))
```
thumbnail, line 30: Why not rethrow this exception? Currently it'll swallow the error.

src/formatting_stack/branch_formatter.clj
```diff
@@ -75,37 +81,60 @@
...
+                                   :overrides overrides
+                                   :formatters (or formatters formatter-factory)
+                                   :linters linter-factory
+                                   :processors processor-factory
                                    :reporter reporter
-                                   :linters linters
                                    :in-background? in-background?)))
 
-(defn lint-branch! [& {:keys [target-branch in-background? reporter]
+(defn lint-branch! [& {:keys [target-branch in-background? reporter overrides]
```
thumbnail, line 130: Why nest formatters/linters/processors/strategies under `override`? They seem direct arguments for `lint-branch!`

src/formatting_stack/core.clj
```diff
@@ -47,30 +48,51 @@
...
+                                                                            ^::protocols.processor/processor-factory processors
+                                                                            third-party-indent-specs
+                                                                            reporter
+                                                                            in-background?]
+                         {formatter-overrides                 :formatters
+                          linter-overrides                    :linters
+                          processor-overrides                 :processors
+                          {formatters-strategies :formatters
+                           linters-strategies    :linters
+                           processors-strategies :processors} :strategies} :overrides}]
```
thumbnail, line 67: Can we deal with the overrides one level higher e.g. create a fn which applies overrides, then call `format!`? 

That way `format!` does less, and doesn't have to change shape, except for the removal of defaults (which is a good thing imo)

src/formatting_stack/core.clj
```diff
@@ -47,30 +48,51 @@
...
            (doall)))))
 
-(defn format! [& {:keys [strategies
-                         third-party-indent-specs
-                         formatters
-                         linters
-                         processors
-                         reporter
-                         in-background?]}]
+(speced/defn format! [& {:keys                                             [^vector? strategies
```
thumbnail, line 51: Not a big fan of the factory-concept here. We could assoc the strategies in `format!` (they are impl detail anyway), which would remove the need of creating the formatters/etc here, retaining the vectors as params. 

`third-party-indent-specs` is still an issue, suggestion posted below

src/formatting_stack/git_status_formatter.clj
```diff
@@ -0,0 +1,135 @@
...
+   [formatting-stack.protocols.processor :as protocols.processor]
+   [formatting-stack.reporters.pretty-printer :as pretty-printer]
+   [formatting-stack.strategies :as strategies]
+   [medley.core :refer [deep-merge]]
+   [nedap.speced.def :as speced]
+   [nedap.utils.modular.api :refer [implement]]))
+
+(speced/defn default-formatters [_
+                                 ^vector? formatters-strategies
+                                 ^map? third-party-indent-specs]
```
thumbnail, line 31: Let's add a default for `third-party-indent-specs` in `cljfmt/new and `cider/new`. Then we can remove the references to `third-party-indent-specs`.

Consumers should overwrite the entire member if they want to change the specs. (which is how other members should work anyway). wdyt?

src/formatting_stack/git_status_formatter.clj
```diff
@@ -0,0 +1,135 @@
...
+              (assoc :strategies cached-strategies))
+          (-> (formatters.how-to-ns/new {})
+              (assoc :strategies cached-strategies))
+          (formatters.no-extra-blank-lines/new)
+          (formatters.newlines/new {})
+          (-> (formatters.trivial-ns-duplicates/new {})
+              (assoc :strategies (conj formatters-strategies
+                                       strategies/files-with-a-namespace
+                                       strategies/exclude-edn)))
+          (when (strategies/refactor-nrepl-available?)
```
thumbnail, line 47: Can we solve this with a override?

src/formatting_stack/branch_formatter.clj
```diff
@@ -75,37 +81,60 @@
...
+    protocols.linter/--linters default-linters))
+
+(def processor-factory
+  (implement {}
+    protocols.processor/--processors default-processors))
+
+(defn format-and-lint-branch!
+  "Note that files that are not completely staged will not be affected.
+
+  You can override that with `:blacklist []`."
```
vemv, line 113: The docstring is correctly reflects the intended behavior, however not _linting_ unstaged files is suboptimal.

After all, linting things is safe, so there's no reason to omit it (and it tends to be surprising; has bit me in the past).

Should be easy to fix.

src/formatting_stack/branch_formatter.clj
vemv: The first thing that comes to mind is that the config should be a single map, so that it can become _data_.

* data (a single hashmap) is simpler than code (a more intrincate invocation consisting of multiple arguments)
* a single hashmap can be (more) easily a resource, a piece of Lein config, etc.
  * this is important, as one will want to share config across projects.
    * I envision this being accomplished by creating a single dep, containing  a .edn as a resource. Zero config or `require`s to be done for consumers: we just _discover_ the presence of such a config file.
* Hashmaps can be (deep) merged
  * Not hard to imagine that people can want a merge of f-s defaults <-> workplace defaults <-> personal defaults.

It's also good to emphasize the notion of overrides as:

* a thing you should be aware of
  * if I can pass `:linters` separately, I can be unaware of the existence a deep-merged overrides system
  * unawareness -> confusion, surprises, etc
* the sole intended API
  * More APIs -> more complexity (surface area for bugs)

src/formatting_stack/branch_formatter.clj
```diff
@@ -75,37 +81,60 @@
...
+  [& {:keys [target-branch in-background? reporter formatters blacklist overrides]
+      :or   {target-branch  "master"
+             in-background? (not (System/getenv "CI"))
+             reporter       default-reporter}}]
```
vemv, line 117: We could :pre that no :linters, :formatters etc are passed as an argument (b/c deprecation)

src/formatting_stack/core.clj
vemv: > We could assoc the strategies in format! (they are impl detail anyway)

See my comment beginning with `The first thing that comes to mind [...]`

src/formatting_stack/protocols/spec.clj
```diff
@@ -22,6 +25,23 @@ so that final users can locate them and configure them."
...
+                                                        (let [ks (->> x
+                                                                      keys ;; fails if (reify)
+                                                                      (filter keyword?)
+                                                                      (filter (fn [x]
+                                                                                (-> x name #{"id"}))))]
+                                                          (map (partial get x) ks)))))
+                                               (apply concat))]
+                                  (when (seq ids)
+                                    (assert (apply distinct? ids)
+                                            "Members should have unique ids"))
```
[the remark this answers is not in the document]
vemv: Will check what can be done. At the very least a `pr-str ids` is due

test/unit/formatting_stack/impl/overrides.clj
vemv: There's an `are` test case that will expect this `::EXCEPTION`, so these aren't omitted

(that's part of why I put these in caps - to make it more immediately visible that something is both declared and consumed)

thumbnail: Could we split that `are` up, so the ones expecting an exception can use `thrown-with-msg?` for a more specific test? (e.g. now _any_ assertionerror will pass)

vemv: Maybe I wouldn't split the `are`- a single `are` means more contrast, less apples-vs-oranges etc

Will look into making this `are` itself more accurate, should be possible

vemv:
Let's try the alternative approach that is in the oven 👀

PR #154, opened 2020-04-07
implement `lint!` for formatters

thumbnail:
## Brief

Run the formatters with lint! tho show the diff/warnings instead of applying them.

Helpful in CI or as dry-run option

## QA plan

1. install `4.3.0-alpha3` in your project
2. run configuration found in `test-resources/linting_formatters_example.clj`
3. assert no files are changed, potentially formatters have show reports on identation, ns forms, etc.

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [ ] Modular design
  * [x] Test coverage
  * [ ] Spec coverage
  * [x] Documentation
  * [ ] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes


src/formatting_stack/formatters/no_extra_blank_lines.clj
```diff
@@ -21,6 +25,30 @@
                                    (spit filename formatted))))))
   nil)
 
+(defn lint! [this files]
+  (letfn [(extra-line-seq [content]
+            (let [matcher        (re-matcher extra-newline-pattern content)
+                  current-offset (fn [] (->> (.start matcher)
+                                             (subs content 0) ;; remove content after cursor
+                                             (re-seq #"\n")
```
thumbnail, line 33:
```suggestion
                                             (string/split-lines)
```

thumbnail:
Right now the diff is not reported, as it tends to clutters the output up _a lot_. It is passed into the report so we can improve on this at any time, and reporters are already extensible for consumers anyway

test/unit/formatting_stack/util.clj
```diff
@@ -15,3 +15,42 @@
...
+     {:begin    30
+      :end      33
+      :filename "b/src/formatting_stack/formatters/no_extra_blank_lines.clj"}
+     {:begin    41
+      :end      41
+      :filename "b/src/formatting_stack/protocols/spec.clj"}]
+
+    "patch with only additions"
+    "test-resources/diffs/4.patch"
+    []))
```
thumbnail, line 56: duplicated case

src/formatting_stack/util.clj
```diff
@@ -141,3 +143,34 @@
...
+            ::rest    (spec/* any?)))
+
+(speced/defn ^coll? diff->line-numbers
+  "Returns maps of consecutive removals/changes (as `:begin` to `:end`) per `:filename` in `diff`."
+  [^string? diff]
+  (->> (io/input-stream (.getBytes diff))
+       (.parse (UnifiedDiffParser.))
+       (data/from-java)
+       (mapcat (fn [{:keys [toFileName hunks]}]
+                 (->> (loop [{::keys [from rest]}
```
vemv, line 159:
```clj
(loop [a
       x
       b y])
```

is inconsistent: better to keeps consistently horizontal or vertical, especially if there's a large `->>` in the middle

src/formatting_stack/util.clj
```diff
@@ -141,3 +143,34 @@
...
+                             result []]
+                        (if (seq rest)
+                          (recur (spec/conform ::grouping-spec rest)
+                                 (conj result from))
+                          (if from
+                            (conj result from)
+                            result)))
+                      (map (fn [grouped-lines]
+                             {:filename toFileName
+                              :begin (apply min (map :lineNumber grouped-lines))
```
vemv, line 175: ->>

test-resources/linting_formatters_example.clj
```diff
@@ -0,0 +1,47 @@
+(ns linting-formatters-example
```
vemv, line 1: Appreciate much the effort, at the same time are these the API and usage patterns we want to foster?

As I see it, with https://github.com/nedap/formatting-stack/pull/152 all usages should become far more concise, involving less Clojure coding and more option passing.

Probably we would have sensible defaults as well for linting in CI as it's a frequent use case.

Maybe we can just remove this .clj, else it's more stuff to maintain / keep track of (particularly as APIs will fluctuate)

test/functional/formatting_stack/formatters/how_to_ns.clj
```diff
@@ -0,0 +1,21 @@
...
+   [clojure.test :refer [are deftest]]
+   [formatting-stack.formatters.how-to-ns :as sut]
+   [formatting-stack.protocols.linter :as linter]
+   [matcher-combinators.test :refer [match?]]))
+
+(deftest lint!
+  (let [linter (sut/new {})]
+    (are [filename expected] (match? expected
+                                     (linter/lint! linter [filename]))
+      "test-resources/invalid_syntax.clj"
```
vemv, line 12: Also cover files without linting warnings?

(same elsewhere)

src/formatting_stack/reporters/impl.clj
```diff
@@ -1,12 +1,13 @@
...
    [nedap.speced.def :as speced]))
 
 (speced/defn normalize-filenames
   "Removes the CWD from the filenames, for more concise output, and also for ensuring correct grouping."
-  [{:keys [^string? filename] :as report}]
-  (assoc report :filename (-> filename
-                              (string/replace (re-pattern (str "^" (System/getProperty "user.dir")))
-                                              "")
-                              (string/replace #"^/" ""))))
+  [{:keys [^::speced/nilable ^string? filename] :as report}]
```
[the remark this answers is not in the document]
thumbnail: It is. Exceptions don't always have a filename. This created a bug where an exception would be swallowed by a spec failure

src/formatting_stack/formatters/no_extra_blank_lines.clj
```diff
@@ -21,6 +22,26 @@
...
+                 (map-indexed (fn [idx line] {:lineNumber idx :line line}))
+                 (partition 2 1)
+                 (filter (fn [[{left :line} {right :line}]] (= "" left right)))
+                 (map (comp inc :lineNumber first))))]
+    (->> files
+         (process-in-parallel! (fn [filename]
+                                 (->> (extra-line-seq (slurp filename))
+                                      (mapv (fn [line]
+                                              {:filename filename
+                                               :msg      "File has extra blank lines"
```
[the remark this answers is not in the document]
thumbnail: Right now it reports on a per-violaton base. So 3 consecutive newlines would trigger 2 warnings. I could try and group those?

src/formatting_stack/formatters/newlines.clj
```diff
@@ -13,8 +17,24 @@
                                  (impl/so-fix-them [filename] :expected-newline-count expected-newline-count)))))
   nil)
 
-(speced/defn new [{:keys [^pos-int? expected-newline-count]
-                   :or   {expected-newline-count 1}}]
+(defn lint! [{:keys [expected-newline-count]} files]
+  (binding [*out* (StringWriter.)
```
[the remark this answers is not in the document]
thumbnail: Because we also bind *err* in `lint!`. Iirc the lib we're using prints the warnings and info to both stderr as stdout when linting (not when formatting). 

I could intercept out and err in both cases though.

src/formatting_stack/formatters/clean_ns.clj
```diff
@@ -57,20 +53,42 @@
...
+                                 (println "Cleaning unused imports:" filename)
+                                 (write-ns-replacement! filename ns-replacement)))))
   nil)
 
+(defn lint! [this files]
+  (->> files
+       (process-in-parallel! (fn [filename]
+                               (when-let [{:keys [final-ns-form-str
+                                                  original-ns-form-str]} (replaceable-ns-form this filename)]
+                                 (let [diff (#'cljfmt.diff/unified-diff filename original-ns-form-str final-ns-form-str)]
```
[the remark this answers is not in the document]
thumbnail: It's a private var. Might want to pull in the entire function though (as we use it in St least 2 places)

project.clj
```diff
@@ -9,10 +9,12 @@
                  [com.nedap.staffing-solutions/utils.collections "2.1.0"]
                  [com.nedap.staffing-solutions/utils.modular "2.2.0-alpha3"]
                  [com.nedap.staffing-solutions/utils.spec.predicates "1.1.0"]
+                 [io.reflectoring.diffparser/diffparser "1.4"]
```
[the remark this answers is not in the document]
thumbnail: We use the diff parser to show where the formatter would've wanted to make the change, so we require parsing the diff for meaningful reporting. I will see if I can change the dep to `com.googlecode.java-diff-utils` (already a transient dep)

test-resources/linting_formatters_example.clj
thumbnail: I was thinking #152 could showcase its effectiveness by changing this file to the new format. Having extra documentation on old vs new configuration. When this PR is delivered I think this should come with some advice on usage.

project.clj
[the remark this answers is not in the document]
thumbnail:
```patch
diff --git a/src/formatting_stack/util.clj b/src/formatting_stack/util.clj
index 8a7df6c..da362f5 100644
--- a/src/formatting_stack/util.clj
+++ b/src/formatting_stack/util.clj
@@ -1,6 +1,6 @@
 (ns formatting-stack.util
   (:require
-   [clojure.java.data :as data]
+   [clojure.java.data :refer [from-java]]
    [clojure.java.io :as io]
    [clojure.spec.alpha :as spec]
    [clojure.tools.namespace.file :as file]
@@ -10,11 +10,13 @@
    [nedap.speced.def :as speced]
    [nedap.utils.collections.eager :refer [partitioning-pmap]]
    [nedap.utils.collections.seq :refer [distribute-evenly-by]]
-   [nedap.utils.spec.predicates :refer [present-string?]])
+   [nedap.utils.spec.predicates :refer [present-string?]]
+   [clojure.string :as str]
+   [clojure.string :as string])
   (:import
    (clojure.lang IBlockingDeref IPending)
-   (io.reflectoring.diffparser.api UnifiedDiffParser)
-   (java.io File)))
+   (java.io File)
+   (difflib DiffUtils)))
 
 (defmacro rcomp
   "Like `comp`, but in reverse order.
@@ -144,33 +146,39 @@
 (defn colorize [s color]
   (str \u001b (ansi-colors color) s \u001b (ansi-colors :reset)))
 
-(spec/def ::grouping-spec
-  (spec/cat ::neutral (spec/* (comp #{"NEUTRAL"} :lineType))
-            ::from    (spec/+ (comp #{"FROM"} :lineType))
-            ::rest    (spec/* any?)))
-
 (speced/defn ^coll? diff->line-numbers
-  "Returns maps of consecutive removals/changes (as `:begin` to `:end`) per `:filename` in `diff`."
+  "Returns maps of consecutive changes (as `:begin` to `:end`) per `:filename` in `diff`."
   [^string? diff]
-  (->> (io/input-stream (.getBytes diff))
-       (.parse (UnifiedDiffParser.))
-       (data/from-java)
-       (mapcat (fn [{:keys [toFileName hunks]}]
-                 (->> (loop [{::keys [from rest]}
-                             (->> hunks
-                                  (mapcat (fn [{:keys [lines] {:keys [lineStart]} :fromFileRange}]
-                                            (->> lines
-                                                 (remove (comp #{"TO"} :lineType))
-                                                 (map-indexed (fn [idx line] (assoc line :lineNumber (+ idx lineStart)))))))
-                                  (spec/conform ::grouping-spec)) ;; group consecutive FROM-changes
-                             result []]
-                        (if (seq rest)
-                          (recur (spec/conform ::grouping-spec rest)
-                                 (conj result from))
-                          (if from
-                            (conj result from)
-                            result)))
-                      (map (fn [grouped-lines]
-                             {:filename toFileName
-                              :begin (apply min (map :lineNumber grouped-lines))
-                              :end (apply max (map :lineNumber grouped-lines))})))))))
+  (->> (string/split diff #"diff --git ")
+       (drop 1) ;; first match is empty
+       (mapcat (speced/fn [^present-string? file-diff]
+                 (speced/let [[_, ^present-string? filename] (re-find #"^a\/(.*) b\/" file-diff)]
+                   (->> (str/split-lines file-diff)
+                        (DiffUtils/parseUnifiedDiff)
+                        (from-java)
+                        (:deltas)
+                        (mapcat (speced/fn [{{^coll? rev :lines}                       :revised
+                                             {:keys [^coll? lines ^pos-int? position]} :original}]
+                                  (let [no-additions (->> rev
+                                                          (filter (fn [idx x]
+                                                                    (= x (get (vec lines) idx)))))]
+                                    (println ::no-additions no-additions)
+                                    (->> lines
+                                         (keep-indexed (fn [idx x]
+                                                         (when-not (= (get (vec no-additions) idx) x)
+                                                           (println ::xxx x)
+                                                           (println ::rev (get (vec no-additions) idx))
+                                                           (+ 1 position idx))))
+                                         (reduce (fn [acc line-number]
+                                                   (let [{:keys [begin end]
+                                                          :or   {begin 0, end 0}} (last acc)]
+                                                     (if (= (inc end) line-number)
+                                                       ;; replace last one if consecutive
+                                                       (conj (butlast acc) {:begin    begin
+                                                                            :end      line-number
+                                                                            :filename filename})
+                                                       ;; add new entry otherwise
+                                                       (conj acc {:begin    line-number
+                                                                  :end      line-number
+                                                                  :filename filename}))))
+                                           [])))))))))))
```

This impl has a bug where it can't detect moved code, because I compare the diff-parts myself instead of the parser.

----

The diffutils-parser can only deal with per-file diffs, so we need to partially parse the diff ourselves in this case. 

Keeping that in mind, and seeing how small and effective the diffparser is, so we'll need to keep it in. We can re-evaluate this choice in the future as it's not public api anyway. Will add a couple more tests.

I found a way around the `::grouping-spec` this way (so this wasn't a total waste of efforts).

src/formatting_stack/formatters/clean_ns.clj
```diff
@@ -57,20 +53,42 @@
...
+                                   (->> (diff->line-numbers diff)
+                                        (mapv (fn [{:keys [begin]}]
```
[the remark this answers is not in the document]
thumbnail: What is your suggestion here? The `diff->line-numbers`-function returns a list of changed sections, for each we need a hashmap. I don't see how we can do that differently

src/formatting_stack/formatters/no_extra_blank_lines.clj
```diff
@@ -21,6 +22,26 @@
                                    (spit filename formatted))))))
   nil)
 
+(defn lint! [this files]
+  (letfn [(extra-line-seq [content]
```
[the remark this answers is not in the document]
thumbnail: I think it's sufficiently covered by the unit-test for `lint!`. It's not public API and essentially the only function in the body of `lint!`.

src/formatting_stack/formatters/how_to_ns.clj
```diff
@@ -21,8 +24,28 @@
...
 
+(defn lint! [{:keys [how-to-ns-options]} files]
+  (->> (remove #(str/ends-with? % ".edn") files)
+       (process-in-parallel! (fn [filename]
+                               (let [contents  (slurp filename)
+                                     formatted (how-to-ns/format-initial-ns-str contents how-to-ns-options)]
+                                 (when-not (= contents formatted)
+                                   (let [diff (#'cljfmt.diff/unified-diff filename contents formatted)]
+                                     (->> (diff->line-numbers diff)
+                                          (mapv (fn [{:keys [begin]}]
```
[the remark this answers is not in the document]
thumbnail: I did this everywhere except for the output of `diff->line-numbers` where i speced the return shape

src/formatting_stack/formatters/clean_ns.clj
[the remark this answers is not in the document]
vemv: Let's put it another way - what do we do if either `diff/unified-diff ` or `diff->line-numbers` throw an exception right here?

It's always possible since Git corner cases are to be expected. Which we wouldn't be particularly prepared against, because:

* `thombergs/diffparser` cannot be considered battle-tested for us (yet)
* `cljfmt` isn't a primarily diff-concerned lib
  * Please imagine what the author's response would be if we found a corner case for something cljfmt does not explicitly support? Or a breaking change?

Needless to say, I'm not against this PR or anything, but enough caution should be placed around the underlying code. Right now it seems the whole linting part revolves around diffs. Seems pretty risky.

src/formatting_stack/formatters/cljfmt.clj
```diff
@@ -14,6 +18,29 @@
...
 
-(speced/defn new [{:keys [third-party-indent-specs] :as options}]
-  (implement (assoc options :id ::id)
-    formatter/--format! format!))
+(defn lint! [{:keys [third-party-indent-specs]} files]
+  (->> files
+       (process-in-parallel! (fn [filename]
+                               (let [indents (impl/cljfmt-indents-for filename third-party-indent-specs)
+                                     {{:keys [okay]} :counts
+                                      :keys [diff]} (#'cljfmt.main/check-one {:indents indents} filename)]
```
vemv, line 26: Please don't invoke private memebers - f-s users can always bump `cljfmt` independently so `cljfmt` is not a dep which we control

src/formatting_stack/formatters/cljfmt.clj
```diff
@@ -14,6 +18,29 @@
...
+                                 (when (zero? okay)
+                                   (->> (diff->line-numbers diff)
+                                        (mapv (fn [{:keys [begin end]}]
+                                                {:filename filename
+                                                 :diff     diff
+                                                 :level    :warning
+                                                 :column   0
+                                                 :line     begin
+                                                 :msg      (str "Indentation is wrong at " (->> (dedupe [begin end])
```
vemv, line 35: s/indentation/whitespace

(cljfmt also formats `a )` -> `a)`)

src/formatting_stack/util.clj
```diff
@@ -141,3 +141,11 @@
 
 (defn colorize [s color]
   (str \u001b (ansi-colors color) s \u001b (ansi-colors :reset)))
+
+(defmacro silence
+  "Execute body without printing to `*out*` or `*err*`"
+  [& body]
```
vemv, line 147: `:style/indent 0`

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,46 @@
...
+   [clojure.java.io :as io]
+   [clojure.spec.alpha :as spec]
+   [clojure.string :as string]
+   [formatting-stack.util :refer [rcomp]]
+   [nedap.speced.def :as speced]
+   [nedap.utils.spec.predicates :refer [present-string?]])
+  (:import
+   (io.reflectoring.diffparser.api UnifiedDiffParser)))
+
+(spec/def ::begin pos-int?)
```
vemv, line 13: `start` or `beginning`?

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,46 @@
...
+(spec/def ::end pos-int?)
+(spec/def ::filename present-string?)
+
+(spec/def ::line-numbers
+  (spec/coll-of (spec/keys :req-un [::begin ::end ::filename])))
+
+(speced/defn ^::line-numbers diff->line-numbers
+  "Returns maps of consecutive removals/changes (as `:begin` to `:end`) per `:filename` in `diff`."
```
vemv, line 21: The docstring is quite spec-like

Most of the info can be elided and/or moved to a def-with-doc elsewhere

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,46 @@
...
+  [^string? diff]
+  (->> (io/input-stream (.getBytes diff))
+       (.parse (UnifiedDiffParser.))
```
vemv, line 24: Please don't use ->> for java interop

Only ->

This style is 100% consistent in formatting-stack, speced.def et al

The intent being to standardize what these can be used for, minimizing mental overhead (i.e. I shouldn't have to thread in my head)

test-resources/linting_formatters_example.clj
vemv: > I was thinking #152 could showcase its effectiveness by changing this file to the new format. 

Effectiveness is better proven by exercising an alpha for a while

In practice we'd introduce more doc that can confuse a casual user for a while. It's not good to introduce churn, e.g constant API changes or patterns of usage.

Otherwise we complect internal and external goals

test/unit/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,47 @@
...
+   [clojure.test :refer [are deftest is testing]]
+   [formatting-stack.util.diff :as sut]))
+
+(deftest diff->line-numbers
+  (are [description filename expected] (testing description
+                                         (is (= expected
+                                                (sut/diff->line-numbers (slurp filename))))
+                                         true)
+
+    "only additions yields empty result"
```
vemv, line 12: Can you make all these `description`s more expressive?

Honestly I have trouble understanding them. I'd suggest using phrasing that describes the use cases from a high level perspective. This emphasizes the notion of tests-as-documentation.

(As a rule of thumb: descs should have a verb)

src/formatting_stack/formatters/clean_ns.clj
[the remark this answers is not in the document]
thumbnail: > Let's put it another way - what do we do if either `diff/unified-diff` or `diff->line-numbers` throw an exception right here?

In both cases it's caught here: https://github.com/nedap/formatting-stack/blob/57e4651fa0913bc4a65415c07b86c3e1dac403a2/src/formatting_stack/util.clj#L99-L103 appended to the result of the report and continues processing.

> `cljfmt` isn't a primarily diff-concerned lib 

All of the formatters are essentially generating a diff and applying that.

Since #163, we _could_ make the diff-parsing opt-in, but it already is. The formatters don't run as linters by default.

I think we need an alpha and some incubation time in some internal projects to see the effects of this PR.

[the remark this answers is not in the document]
vemv: > In both cases it's caught here:

Can be good enough, OTOH hitting errors (i.e. an increased error rate) is not desirable, even if those errors are caught and reported.

We can increase confidence around this feature by assessing the `.patch` files that are part of this PR:

**1.-** Do those cover all cases?

* Completetly added file
* Completetly removed file
* Merge conflicts
* Renames
* Other?
  * Worth potentially digging into other projects' test suites

**2.-** Those .patch files would be ideally also exercised against `cljfmt.diff/unified-diff`.

Normally we shouldn't craft tests for _other_ libs, but:

* cljfmt is not a lib that promises to provide an API for diffs
* we're using a defn that is public mostly by chance.
  * So we should create our own wrapper that invokes `cljfmt.diff/unified-diff`
  * WIth this wrapping, it becomes more natural that we craft tests for it.
  * Maybe: vendorize this defn if after asking, we verify that this indeed isn't quite a public API.

> All of the formatters are essentially generating a diff and applying that.

Doesn't seem true to me. Formatters spit contents to filenames. I don't know of a single formatter that emits a .patch (or even: that computes a fine-grained difference at all) and applies it to a file somehow.

The fact that formatting can be _seen_ as patch applications seems orthogonal to the inner workings of these formatters.

(there's a difference between computing diffs after-the-fact, and formatting things out of diffs)

> The formatters don't run as linters by default.

True, but using formatters-as-linters is something various parts will want soonish, so it can be assumed that these will be enabled. And when that happens, builds shouldn't start faiiling. 

> I think we need an alpha and some incubation time in some internal projects to see the effects of this PR.

Agreed, seems quite a precondition to me but also we'd gain a lot of confidence if you could confirm a bit more thoroughtly that we've explored this problem space well, and that we're shipping all reasonable tests, cautions, etc.

test-resources/linting_formatters_example.clj
thumbnail: I added an explanation in the docstring that this config is subject to change. I want to keep the example in, so we have some guidance until #152 or a similar solution is in.

src/formatting_stack/formatters/clean_ns.clj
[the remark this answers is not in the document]
thumbnail: Thought about this a bit, Will keep the current solution (computing diffs, reporting on the patches).

To increase confidence I extracted the unified-diff function, and added tests for the current usecases. They can be extended whenever we need more.

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,69 @@
...
+(spec/def ::start pos-int?)
+(spec/def ::end pos-int?)
+(spec/def ::filename present-string?)
+
+(speced/def-with-doc ::line-numbers
+  "maps of consecutive removals/changes (as `:start` to `:end`) per `:filename`."
+  (spec/coll-of (spec/keys :req-un [::start ::end ::filename])))
+
+(speced/defn ^::line-numbers diff->line-numbers
+  [^string? diff]
```
vemv, line 25: >  `[^string? diff]`

It seems important to spec the first line ,such as:

```
diff --git a/test/unit/formatting_stack/util.clj b/test/unit/formatting_stack/util.clj
```

* left and right hand should refer to existing files
  * guards us against race conditions, or various plausible bugs
* left and right hand should probably have a "normalized" form 
  * relative or absolute filepaths, but not something unspecified
* Can `/dev/null` appear?
  * often does in git diffs

test/unit/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,78 @@
+(ns unit.formatting-stack.util.diff
+  (:require
+   [clojure.test :refer [are deftest is testing use-fixtures]]
+   [formatting-stack.test-helpers :refer [with-mocked-diff-path]]
+   [formatting-stack.util.diff :as sut]))
+
+(use-fixtures :once with-mocked-diff-path)
+
+(deftest diff->line-numbers
```
vemv, line 9: checking out the files 1.patch to 4.patch, I think we could benefit from extra coverage via:

* diffs where there's /dev/null in either left or right hand
* diffs denoting a file renaming

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,69 @@
...
+  (->> (.parse (UnifiedDiffParser.) (io/input-stream (.getBytes diff)))
```
vemv, line 26: Perhaps as a measure of caution, we can wrap this code with a try/catch.

In the catch clause, we return a `::line-numbers` vector with size 1; its object would have line 0 and column 0 but the actual filename parsed from the `diff` argument.

This way, linters always report something even if diff parsing went wrong.

Point being, f-s users benefit more from a less expressive linting report ("line 0 column 0") than from an extraneous stacktrace.

Of course, we can also print that something went wrong so that these aren't swallowed opaquely.

---

(especially if doing this, please add :min-count 1 to the ::line-numbers spec)

src/formatting_stack/formatters/how_to_ns.clj
```diff
@@ -16,13 +19,33 @@
...
+                                 (when-not (= contents formatted)
+                                   (let [diff (diff/unified-diff filename contents formatted)]
+                                     (->> (diff->line-numbers diff)
+                                          (mapv (fn [{:keys [start]}]
+                                                  {:filename filename
+                                                   :diff diff
+                                                   :line start
+                                                   :column 0
+                                                   :level :warning
+                                                   :msg "Badly formatted namespace"
```
vemv, line 41: Better: `Badly formatted ns form`, since `namespace` can also denote the whole file.

Also maybe I'd try to avoid `Badly` since it sounds judgemental. Using `:warning-details-url "https://stuartsierra.com/2016/clojure-how-to-ns.html"` can compensate

src/formatting_stack/formatters/newlines.clj
```diff
@@ -1,20 +1,36 @@
...
    [formatting-stack.protocols.formatter :as formatter]
-   [formatting-stack.util :refer [process-in-parallel!]]
+   [formatting-stack.protocols.linter :as linter]
+   [formatting-stack.util :refer [process-in-parallel! silence]]
    [nedap.speced.def :as speced]
    [nedap.utils.modular.api :refer [implement]]))
 
 (defn format! [{:keys [expected-newline-count]} files]
-  (with-out-str ;; Supress "All newlines are good, nothing to fix."
+  (silence ;; Supress "All newlines are good, nothing to fix."
```
vemv, line 12: Maybe for both instances of `silence` in this pr, they'd be better placed around `impl/*`, and no broader.

That way it's clearer which part is at fault. Also I'd be slightly afraid of parallelism mixed with silencing (it has worked well for us, but seems delicate)

src/formatting_stack/formatters/no_extra_blank_lines.clj
```diff
@@ -21,6 +22,29 @@
                                    (spit filename formatted))))))
   nil)
 
+(defn lint! [this files]
+  (->> files
+       (process-in-parallel! (fn [filename]
```
vemv, line 27: This logic could be made pure by decoupling it from `filename`. That way, you could unit-test it, and extensively.

It'd be particularly important that formatting and linting tests shared the same exact corpus in `unit.formatting-stack.formatters.no-extra-blank-lines`,  that way we prove they behave consistently between them.

test/formatting_stack/test_helpers.clj
```diff
@@ -8,3 +9,10 @@
   (str "file:" (-> filename
                    File.
                    .getAbsolutePath)))
+
+(defn with-mocked-diff-path
+  "Fixture to stub the absolute path in #'util.diff/unified-diff"
+  [t]
+  (with-redefs [util.diff/to-absolute-path (fn [filename]
```
vemv, line 16: Would be nice to keep the codebase `with-redefs`-free

Maybe `unified-diff` could look like this:

```clj
(def *to-absolute-path-fn* to-absolute-path)

(speced/defn ^string? unified-diff
  "derives a patch derived from the original and revised file contents in a Unified Diff format"
  ([^string? filename ^string? original ^string? revised]
   (letfn [(lines [s]
             (string/split s #"\n"))
           (unlines [ss]
             (string/join "\n" ss))]
     (unlines (DiffUtils/generateUnifiedDiff
               (->> filename *to-absolute-path-fn* (str "a"))
               (->> filename *to-absolute-path-fn* (str "b"))
...
```

src/formatting_stack/util/diff.clj
thumbnail: > It seems important to spec the first line

I don't see the importance. The `UnifiedDiffParser#parse` method literally parses the diff (and will throw `IllegalStateException` if it runs into an error) which is far more extensive and expressive than a regex could be.

> left and right hand should refer to existing files

> left and right hand should probably have a "normalized" form

Seems overly tight. A diff does not need to be accessible in order be parsable. arbitrary restrictions like these can come as a surprise in the future.

If a consumer requires the path to be accessible, it can specify it itself.

src/formatting_stack/util/diff.clj
thumbnail: > Perhaps as a measure of caution, we can wrap this code with a try/catch.

Seems like a good idea. However, multiple filenames can be present in 1 diff, and if parsing fails we'd still need some kind of backup parsing. 

Pulling the safe-guard one level higher, such as here: 
https://github.com/nedap/formatting-stack/blob/3360e801c4f34c8c4672bf857dcec274982546e6/src/formatting_stack/formatters/clean_ns.clj#L80

Except it's already wrapped in a try-catch by `process-in-parallel!`, which neatly reports a `:formatting-stack/report-processing-error`.

I think we should simply alpha this for a while to gain more confidence in this approach.

> (especially if doing this, please add :min-count 1 to the ::line-numbers spec)

A diff has no guarantee to contain any removed lines. an empty collection is a valid return value.

src/formatting_stack/formatters/how_to_ns.clj
thumbnail: Good call, went with `"Detected unsorted, renamed or extra clauses in the ns format"` and the url 👍

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,70 @@
...
+(spec/def ::end nat-int?)
+(spec/def ::filename present-string?)
+
+(speced/def-with-doc ::line-numbers
+  "maps of consecutive removals/changes (as `:start` to `:end`) per `:filename`."
+  (spec/coll-of (spec/keys :req-un [::start ::end ::filename])))
+
+(speced/defn ^::line-numbers diff->line-numbers
+  [^string? diff]
+  (->> (.parse (UnifiedDiffParser.) (io/input-stream (.getBytes diff)))
```
vemv, line 26: > I don't see the importance.

Ensuring that we are performing a valid diff comparison seems better placed as a responsiblity of f-s. Third party libs can be swapped out, have bugs, evolve and lack a precondition system.

It's quite plausible that any of these libs could decide to return an empty diff on invalid input (e.g. nonexisting file, mixed relatve vs. absolute path).

Specs or tests cover us against these in a precise fashion beyond what fine-grained, one-off reasoning can accomplish.

> [...] a regex could be.

No need for regex if that's part of the perceived problem

> A diff does not need to be accessible in order be parsable.

Forbidding non-accessible files rules out plausible bugs

> arbitrary restrictions like these can come as a surprise in the future.

It'd be a good kind of surprise because it would mean that there are corner cases that lack test/spec coverage. It's better than other plausible surprises (e.g. an opaquely wrong result that didn't throw an exception)

src/formatting_stack/util/diff.clj
vemv: > Pulling the safe-guard one level higher, such as here

Needless to say, it'd be better to have a single, abstract point for logic instead of repeating it around. The `UnifiedDiffParser` is the part that can plausibly fail, a try/catch should be as close to it as possible.

> I think we should simply alpha this for a while to gain more confidence in this approach.

As you can understand, an alpha cannot prove the lack of issues, particularly when we work in different teams. I'd be more comfortable with a try/catch; if you aren't, why don't we use a feature flag?

Example:

```clj
(defn diff->line-numbers* ...) ;; old impl, unaltered

(def diff->line-numbers
  (cond-> diff->line-numbers
    (System/getProperty "formatting-stack.diff.try") wrap-try))
```

vemv (commented):
Thanks for the iterations 💪

src/formatting_stack/util/diff.clj
vemv: > A diff has no guarantee to contain any removed lines. an empty collection is a valid return value.

My reasoning was: if one is invoking `diff->line-numbers` in the first place, it's because some linter has something to report. But if the return value is `[]`, then nothing is being reported.

WDYT?

thumbnail: My main concern with the try-catch is that it should live inside the linter in order to have enough context to create a meaningful fallback-result. `diff->line-numbers` should throw an exception on invalid diffs (might need to reflect it in the docstring, so consumers can expect it).

> if one is invoking diff->line-numbers in the first place, it's because some linter has something to report.

In our current usage: yes. I think this makes sense but be live one level higher. The linter should warn / error if a diff was generated, but no line-numbers could be determined.

src/formatting_stack/formatters/no_extra_blank_lines.clj
```diff
@@ -21,6 +24,34 @@
                                    (spit filename formatted))))))
   nil)
 
+(speced/defn ^{::speced/spec (spec/coll-of pos-int?)} report-extra-newlines [^string? s]
```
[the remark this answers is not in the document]
thumbnail: > not very desirable to keep two unrelated impls

They do two different tasks. One replaces consecutive newlines, the other reports. They're not exactly interchangable. I dislike the symmetry, but the simplicity of `without-extra-newlines` shouldn't be replaced by a more complex implementation as a goal. 

There's tests to assert similar the behaviour of both.


If we really want symmetry between lint! and format!, i'd opt for generating a diff and using the reported line-numbers.

test-resources/diffs/5.patch
```diff
@@ -0,0 +1,4 @@
+diff --git a/src/formatting_stack/linters/line_length.clj b/src/formatting_stack/linters/line_length2.clj
+similarity index 100%
```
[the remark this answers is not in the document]
thumbnail: The test-suite is adequate right now. If any bugs pop up we can extend it. Having a test-suite testing every nook and cranny of git / patches is not a goal.

src/formatting_stack/util/diff.clj
vemv: > I think this makes sense but be live one level higher.

Thanks, yes I suspected this was the case. Would be happy for any kind of approach

src/formatting_stack/formatters/no_extra_blank_lines.clj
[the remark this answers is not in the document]
vemv: > They do two different tasks.

Ok

It's fairly unfortunate that we can't share formatting and linting impls automatically, this has been the case for every 'member' in this PR. Admittedly it's not easy.

We could keep an eye on this and might able to fix it at some point.

test-resources/diffs/5.patch
```diff
@@ -0,0 +1,4 @@
...
+rename from src/formatting_stack/linters/line_length.clj
```
vemv, line 3: > The test-suite is adequate right now. 

> Having a test-suite testing every nook and cranny of git / patches is not a goal.

It's useful to remember the monorepo / git-status issues we had at some point; it took us a few iterations to get the solution right. Exhaustive testing+specing helped us gain confidence, once we had that in place we never saw an issue again.

More generally it's fairly uncontroversial that spec+tests precision correlate directly to bug-free software. They're one of the few automated tools we have directly at hand here and I'd appreciate their continued use just like every other feature in f-s.

This bug-freedom is particularly important for something that can be easily run as a blocking CI check.

src/formatting_stack/util/diff.clj
thumbnail: Still not convinced. This function derives line numbers from a diff. the filepaths are part of the data and do not need to be related to the local filesystem.

If the existence is required, we should put it in the consumer of this fn. 

This function is currently pure (adding filesystem checking would make it impure), and i'd like to keep it that way.

test-resources/diffs/5.patch
thumbnail: I'm not advocating against tests or questioning their usefulness in general.

We should be careful not to create numerous redundant test cases, they can hurt the clarity in case of an error or even a false sense of safety.

This function is not public api, and is only used by cljfmt, trivial-ns-duplicates, how-to-ns, clean-ns. Which is why the current test-suite is adequate.

src/formatting_stack/util/diff.clj
vemv: > This function is currently pure (adding filesystem checking would make it impure), and i'd like to keep it that way.

Ok, let's leave it as-is

test-resources/diffs/5.patch
vemv: These aren't redundant tests because there's a very limited set of cases: 8 (1-7.patch to the proposed 8.patch). Each of them is _different_.

It's not like we're adding random tests that don't exercise any aspect in particular (we can call this _quantitative safety_). Instead, each test adds distinct coverage (qualitative safety).

When all known cases are covered (1-8), we get _exhaustive_ coverage. In this case we either have exhaustive coverage or not: it's a binary thing. Leaving the test suite "open" (non-exhaustive) opens the door to undetected regressions in a future; contrariwise a closed system is categorically superior.

A good example of exhaustiveness being deemed superior (because it rules out undefined behavior) can be found in the [exhaustive pattern matching](https://fsharpforfunandprofit.com/posts/correctness-exhaustive-pattern-matching/) construct, present in various FP langs.

vemv:
Out of curiosity, I ran this code to exercise `diff->line-numbers` over f-s' history itself:

```clj
(let [lines (-> (clojure.java.shell/sh "git" "log" "-p" "--format=") :out clojure.string/split-lines)
      [chunks wip] (->> lines
                        (reduce (fn [[chunks wip] line]
                                  (if (clojure.string/starts-with? line "diff --git a/")
                                    [(conj chunks wip) [line]]
                                    [chunks (conj wip line)]))
                                [[] []]))
      all (conj chunks wip)]
  (->> all
       (map (partial clojure.string/join \newline))
       (filter seq)
       (run! (fn [diff]
               (println (try
                          (formatting-stack.util.diff/diff->line-numbers diff)
                          (catch Exception e
                            diff)))))))
```

Thankfully it only errors on diffs showing a .patch file, which sounds like a tricky case for the underlying Java lib (and a irrelevant case for us: we filter those since only `.clj`-like files are processed)

Still I'd recommend to run this code over other large codebases, as some QAing.

test-resources/diffs/5.patch
thumbnail: settled by https://github.com/nedap/formatting-stack/pull/154#issuecomment-750261251

src/formatting_stack/util/diff.clj
thumbnail: `diff->line-numbers` is always used inside `process-in-parallel!`, so I added a test which verifies those exceptions are caught :).

vemv:
I found out that if  `formatting-stack.util.diff/diff->line-numbers` is passed a string ending in literally `\ No newline at end of file`, it will fail.

[Examples](https://grep.app/search?q=No%20newline%20at%20end%20of%20file)

Stripping it sounds good enough :)

thumbnail:
> I found out that if formatting-stack.util.diff/diff->line-numbers is passed a string ending in literally \ No newline at end of file, it will fail.

I added a test, but couldn't reproduce. Did find a bug when changing the first line of the file, the spec was too strict (`pos-int?` instead of `nat-int?`).

vemv:
Thanks!

Here's a repro over the alpha:

```
(formatting-stack.util.diff/diff->line-numbers "diff --git a/foo.clj b/foo.clj
new file mode 100644
index 000000000..fd88b7b54
--- /dev/null
+++ b/foo.clj
@@ -0,0 +1,9 @@
+    true foo))
\\ No newline at end of file")
```

The error is in the spec, you describe. Although I'm not sure your fix was ideal: by stripping `\\ No newline at end of file`, I don't get a spec error.

i.e. stripping this ending allows us to keep a stronger spec.

WDYT?

(The stronger spec didn't fail when running the "QA snippet" over a large codebase I have at hand)

thumbnail:
The spec was wrong: `:fromFileRange` contains `:lineStart` (can be 0, speced as `nat-int?`), `map-indexed` starts with an `idx` at 0. `(+ 0 0)` returns 0, so it's still a natural int, not a guaranteed positive int.

vemv:
Got it 👍

Maybe you can update the inline spec to match https://github.com/nedap/formatting-stack/blob/923525048bc846206c4f0e835f92fad6725061cd/src/formatting_stack/protocols/spec.clj#L37

vemv:
We can use `[clj-kondo "2021.01.20"]` - per my trying it doesn't bring anything off

Its new `:unresolved-var` option, though possibly handy for people using it with IDE integration, should be turned off in f-s given it's redundant given our `(refresh)` assumption.

(Its related discussion acknowledges that macro-defined vars can bring false postives)

test-resources/linting_formatters_example.clj
```diff
@@ -0,0 +1,49 @@
...
+  "Documents how formatters can be used as linters for a non-mutative experience.
+
+  Not a part of the test suite.
+
+  NOTE: Highly subject to change! see https://github.com/nedap/formatting-stack/pull/152
```
tcoenraad, line 6: I cannot clearly see the connection between #152 and the warning itself as `overrides` is not used

test-resources/ns_unordered.clj
```diff
@@ -0,0 +1,4 @@
+(ns ns-unordered
+  (:require
+   [clojure.string :as foo]
+   [clojure.stacktrace :as bar]))
```
tcoenraad, line 4: Alternative: use visual distinctive namespaces so you can spot the difference in one blink of an eye

test-resources/diffs/2.diff
```diff
@@ -0,0 +1,43 @@
+diff --git a/src/formatting_stack/formatters/trivial_ns_duplicates.clj b/src/formatting_stack/formatters/trivial_ns_duplicates.clj
```
tcoenraad, line 1: Are you able to have this (& following) to be a bit smaller, link 1.diff?

tcoenraad (commented):
I'll test the alpha later this day

test-resources/diffs/2.diff
thumbnail: why?

test-resources/linting_formatters_example.clj
thumbnail: #152 introduces `:overrides`, so it cannot be used. That's the point.


Once `:overrides` is merged, this example shall be updated to use it.

src/formatting_stack/formatters/clean_ns.clj
```diff
@@ -57,20 +53,42 @@
...
+                                                  original-ns-form-str]} (replaceable-ns-form this filename)]
+                                 (let [diff (diff/unified-diff filename original-ns-form-str final-ns-form-str)]
+                                   (->> (diff->line-numbers diff)
+                                        (mapv (fn [{:keys [start]}]
+                                                {:filename filename
+                                                 :diff diff
+                                                 :level :warning
+                                                 :column 0
+                                                 :line start
+                                                 :msg "ns can be cleaned"
```
jwkoelewijn, line 87:
```suggestion
                                                 :msg "ns has unused imports and can be cleaned"
```

src/formatting_stack/util/diff.clj
```diff
@@ -0,0 +1,71 @@
...
+
+                                    :else
+                                    (conj ret {:filename (string/replace-first toFileName "b/" "")
+                                               :start lineNumber
+                                               :end   lineNumber}))))
+                              []))))))
+
+(def diff-context-size 3)
+
+(def ^:dynamic *to-absolute-path-fn*
```
jwkoelewijn, line 53: This could use a small docstring explaining why it is `^:dynamic` and how that is used (figured out this is useful in tests, only by reviewing all of these changes, will be harder once this is actually part of the codebase

jwkoelewijn (approved):
Nice one! Quite extensive test coverage as well, kudos!

src/formatting_stack/formatters/clean_ns.clj
thumbnail: The ns _might_ have unused imports, but it could also be unused requires, refers or be (partly) unordered. I struggle to find a good description of everything for this message.

jwkoelewijn: Check. I was basing my suggestion on the `println` in the `format!` function. Maybe my point is more to at least have the two consistent :)

But yeah, I see how my suggestion is misleading/painting an incomplete picture as well.

test-resources/linting_formatters_example.clj
tcoenraad: Ah, it was not merged yet! It was that old, I expected that it was merged.

test-resources/diffs/2.diff
tcoenraad: To make it more clear what a test makes (or brakes)

thumbnail: The goal is to parse arbitrary diff files. I consider the *.diff files to be opaque / external input. Making them more readable is not really a concern. But we could discuss this tomorrow :).

tcoenraad (approved):
QA'd on Key (using JDK11 btw). Worked just fine 🆗

PR #160, opened 2020-05-05
Exception reporter bug

thumbnail:
## Brief

discovered in #154, in some cases an exception occur, which are passed into reporter, without filenames. This is expected as the exception sometimes escapes the context where the filename is available (especially in `process-in-parallel!`).

This triggers a spec-error which hides the initial exception, as shown in the failing CI here: https://circleci.com/gh/nedap/formatting-stack/1818

## QA plan

Fixed CI after `3a9c4a7` (revert to see breakage)

## Author checklist

* [ ] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/reporters/impl.clj
+++ b/src/formatting_stack/reporters/impl.clj
@@ -1,12 +1,13 @@
 (ns formatting-stack.reporters.impl
   (:require
    [clojure.string :as string]
+   [medley.core :refer [assoc-some]]
    [nedap.speced.def :as speced]))
 
 (speced/defn normalize-filenames
   "Removes the CWD from the filenames, for more concise output, and also for ensuring correct grouping."
-  [{:keys [^string? filename] :as report}]
-  (assoc report :filename (-> filename
-                              (string/replace (re-pattern (str "^" (System/getProperty "user.dir")))
-                                              "")
-                              (string/replace #"^/" ""))))
+  [{:keys [^::speced/nilable ^string? filename] :as report}]
+  (assoc-some report :filename (some-> filename
+                                       (string/replace (re-pattern (str "^" (System/getProperty "user.dir")))
+                                                       "")
+                                       (string/replace #"^/" ""))))
--- a/src/formatting_stack/reporters/impl.clj
+++ b/src/formatting_stack/reporters/impl.clj
@@ -1,12 +1,18 @@
 (ns formatting-stack.reporters.impl
   (:require
+   [clojure.spec.alpha :as spec]
    [clojure.string :as string]
+   [formatting-stack.protocols.spec :as protocols.spec]
    [medley.core :refer [assoc-some]]
    [nedap.speced.def :as speced]))
 
+(speced/def-with-doc ::filename
+  "Filename which can be nil when an exception occurred"
+  (spec/nilable ::protocols.spec/filename))
+
 (speced/defn normalize-filenames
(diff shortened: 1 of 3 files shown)


src/formatting_stack/reporters/impl.clj
```diff
@@ -1,12 +1,13 @@
...
    [nedap.speced.def :as speced]))
 
 (speced/defn normalize-filenames
   "Removes the CWD from the filenames, for more concise output, and also for ensuring correct grouping."
-  [{:keys [^string? filename] :as report}]
-  (assoc report :filename (-> filename
-                              (string/replace (re-pattern (str "^" (System/getProperty "user.dir")))
-                                              "")
-                              (string/replace #"^/" ""))))
+  [{:keys [^::speced/nilable ^string? filename] :as report}]
```
vemv, line 9: wdyt of `def-with-doc`ing `^::speced/nilable ^string?` in a centralized spec?

That way we document the nuance.

It's also a good opportunity to review (and possibly upgrade) other specings of `filename`

test/functional/formatting_stack/core.clj
```diff
@@ -0,0 +1,37 @@
...
+  (throw (ex-info "Kaboom!" {})))
+
+(def failing-formatter
+  (implement {:id ::failing-formatter}
+    formatter/--format! throw-exception))
+
+(defn store-in-latest [{::keys [latest]} args]
+  (reset! latest args))
+
+(defn recording-recorder [latest]
```
vemv, line 21: recording-reporter

test/functional/formatting_stack/core.clj
```diff
@@ -0,0 +1,37 @@
...
+
+(deftest format!
+  (testing "exceptions are passed to the reporter"
+    (let [x (atom nil)]
+      (sut/format! :formatters [failing-formatter]
+                   :linters []
+                   :processors []
+                   :reporter (recording-recorder x)
+                   :in-background? false)
+      (is (match? [{:source    :formatting-stack/process!
```
vemv, line 33: `u.t/expect`?

It would _look_ like overkill atm, but one never knows how tests could grow

src/formatting_stack/reporters/impl.clj
thumbnail: `:formatting-stack.protocols.spec/report` has 2 specs, one where `:filename` is required, and one for exceptions where `:filename` is optional, so the it _is_ documented. But it would be fine in this case to remind the reader.

vemv: >  it is documented

But ` ^::speced/nilable ^string? filename` does not relate to those 2 specs (b/c it's inline), so one cannot necessarily guess that relationship

test/functional/formatting_stack/core.clj
thumbnail: I tried `u.t/expect`, but it's a bit hard to use in this case, as it uses `meta=`. Kept the test as is, can reevaluate when we expand it.

lennartbuit (approved):
Not intimately familiar with the code, but seems good to me

PR #166, opened 2020-08-17
do not calculate classpath-cache on require

thumbnail:
## Brief

This is reported to cause a lot of io, which could cause timeouts when when booting. By putting this into a delay the classpath-cache is calculated on the first kondo linter run.

## QA plan

 1. release alpha
 1. remove `.clj-kondo`-folder from projects known to cause this issue
 1. boot repl; observe no timeout
 1. run f-s; observe creation of cache
 1. run f-s again; observe faster run

## Author checklist

* [ ] The build passes
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/worker/formatting_stack/kondo_classpath_cache.clj
+++ b/worker/formatting_stack/kondo_classpath_cache.clj
@@ -17,7 +17,7 @@
 (def cache-dir (str cache-parent-dir File/separator cache-subdir))
 
 (def classpath-cache
-  (future
+  (delay
     (let [files (-> (System/getProperty "java.class.path")
                     (string/split #"\:"))]
       (-> (File. cache-parent-dir cache-subdir) .mkdirs)
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ As you can see in the screenshot, **formatting-stack** presents linters' outputs
 #### Coordinates
 
 ```clojure
-[formatting-stack "4.2.3"]
+[formatting-stack "4.2.4-alpha1"]
 ```
 
 **Also** you might have to add the [refactor-nrepl](https://github.com/clojure-emacs/refactor-nrepl) dependency.
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
 ;; Please don't bump the library version by hand - use ci.release-workflow instead.
-(defproject formatting-stack "4.2.3"
+(defproject formatting-stack "4.2.4-alpha1"
   ;; Please keep the dependencies sorted a-z.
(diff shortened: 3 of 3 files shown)


worker/formatting_stack/kondo_classpath_cache.clj
```diff
@@ -17,7 +17,7 @@
 (def cache-dir (str cache-parent-dir File/separator cache-subdir))
 
 (def classpath-cache
-  (future
+  (delay
```
vemv, line 20: I'd suggest a `(delay (future x))` so that one gains some responsiveness in some cases.

...Adding comments for both constructs so that we don't forget these intricacies

vemv: Maybe I'd be a bit more comfortable with not losing the responsiveness gain, particularly given this is a problem that can be avoided by setting a generous `{:repl-options {:timeout` (which IME tends to be necessary anyway for other reasons in large projects).

FWIW I have never had the described problem

How about a macro that does: `(if (System/getProperty ...) (future ...) (delay ...))`

I would probably have no problem defaulting to slowness, but one should be able to not lose perf

thumbnail: Sure thing, we're just testing this approach right now.

 Having said that; my personal lein-profile has a repl-timeout of several minutes; and I still experienced timeouts on first boot. 
 - This could be mitigated by not requiring f-s. 
 - Another pointer was that the kondo cache thread would be using a fair amount of cpu time; competing with the repl class loader.
 - retrying to load the repl would result in a booting repl at some point, once the cache was (mostly) created.

I think in hindsight that creating a cache (and thus reading analyzing file on the classpath) upon require is a rather expensive approach. 


---

About the `(delay (future ..` suggestion; i don't see any value in that, could you elaborate on the `responsiveness in some cases` ?. Consumers could still wrap the dereferencing in a future if they'd want to calculate the cache on require.

thumbnail:
mitigated by not loading f-s when the repl is loaded. instead we create aliases which use `requiring-resolve` to load f-s upon first reload / format.

PR #149, opened 2020-03-15
Introduce Eastwood linter parallelization, opt-in

vemv:
## Brief

Introduces Eastwood linter parallelization under a JVM flag (_not_ to be particularly documented).

Although we think the linters are thread-safe, we should take some cautions for catching any issues on time.

See: https://github.com/jonase/eastwood/pull/340

### 1

` (dotimes [i 3] (time (formatting-stack.project-formatter/lint-project!)))` over f-s itself.

Improvement: 17s -> 14s.

> Improvement is larger (e.g. 10 seconds - just verified) in legacy projects where a single given ns can span e.g. 1000LOC.

### 2

CircleCI setup has been expanded. Let's see if it introduce any flakiness over the following couple weeks.

### 3 

Let's set the JVM flag in our personal Lein profiles, to exercise this often.

## Author checklist

* [x] I have QAed the functionality
* [x] The PR has a reasonably reviewable size and a meaningful commit history
* [x] I have run the [branch formatter](https://github.com/nedap/formatting-stack/blob/332a419034ab46fad526a5592f4257353bd695b6/src/formatting_stack/branch_formatter.clj) and observed no new/significative warnings
* [x] The build passes
* [x] I have self-reviewed the PR prior to assignment
* Additionally, I have code-reviewed iteratively the PR considering the following aspects in isolation:
  * [x] Correctness
  * [x] Robustness (red paths, failure handling etc)
  * [x] Modular design
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes
  * [x] Cross-compatibility (Clojure/ClojureScript)

## Reviewer checklist

* [ ] I have checked out this branch and reviewed it locally, running it
* [ ] I have QAed the functionality
* [ ] I have reviewed the PR
* Additionally, I have code-reviewed iteratively the PR considering the following aspects in isolation:
  * [ ] Correctness
  * [ ] Robustness (red paths, failure handling etc)
  * [ ] Modular design
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes
  * [ ] Cross-compatibility (Clojure/ClojureScript)

PR #173, opened 2021-02-23
use :naive parallelization for eastwood

thumbnail:
## Brief

Sets eastwoods `parallel?` to `:naive` for gains ⚡

## QA plan

Run the following snippet in projects with both v4.3.0 and a local install with this change
```clj
(require '[formatting-stack.strategies :as strategies]
         '[formatting-stack.linters.eastwood :as linters.eastwood]
         '[formatting-stack.protocols.linter :as protocols.linter])
(time
  (protocols.linter/lint!
    (linters.eastwood/new {})
    (strategies/all-files :files [])))
```

for formatting-stack this yields a 2x improvement (39s -> 17s)

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [x] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -44,14 +44,16 @@
     (->> @reports
          :warnings
          (map :warn-data)
-         (map (fn [{:keys [uri-or-file-name linter] :strs [warning-details-url] :as m}]
-                (assoc-some m
-                            :level               :warning
-                            :source              (keyword "eastwood" (name linter))
-                            :warning-details-url warning-details-url
-                            :filename            (if (string? uri-or-file-name)
-                                                   uri-or-file-name
-                                                   (-> ^File uri-or-file-name .getCanonicalPath)))))
+         (map (fn [{:keys [uri-or-file-name msg line column linter] :strs [warning-details-url]}]
+                (assoc-some {:level         :warning
+                             :source        (keyword "eastwood" (name linter))
+                             :filename      (if (string? uri-or-file-name)
+                                              uri-or-file-name
+                                              (-> ^File uri-or-file-name .getCanonicalPath))
+                             :column        column
+                             :msg           msg
+                             :line          line}
+                            :warning-details-url warning-details-url)))
          (concat (impl/warnings->reports output)
                  (impl/exceptions->reports @exceptions)))))
 
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -12,12 +12,11 @@
    (java.io File)))
 
 (def default-eastwood-options
-  ;; Avoid false positives or undesired checks:
-  (let [linters (remove #{:suspicious-test :unused-ret-vals :constant-test :wrong-tag}
-                        eastwood.lint/default-linters)]
-    (-> eastwood.lint/default-opts
-        (assoc :linters linters
-               :rethrow-exceptions? true))))
+  (assoc eastwood.lint/default-opts
+         ;; Avoid false positives or undesired checks:
+         :exclude-linters     #{:suspicious-test :unused-ret-vals :constant-test :wrong-tag}
+         :parallel?           :naive
+         :rethrow-exceptions? true))
 
 (def parallelize-linters? (System/getProperty "formatting-stack.eastwood.parallelize-linters"))
 


vemv:
Please verify exactly what Eastwood :parallel? does - I suspect it's dangerous. I've had a negative experience with parallel application of tools.analyzer:

https://github.com/clojure-emacs/refactor-nrepl/pull/247#issuecomment-561009122

thumbnail:
> Please verify exactly what Eastwood :parallel? does 

This change has been verified and there's a QA which has seen extensive testing locally. the :parallel? option parallelizes after the analysis is done. I'd've opted for a partitioning-pmap solution if `lint-namespace` was public API.

vemv:
Ok, I'll state the facts myself:

* `:parallel?` causes a parallel invocation of `lint-namespace` https://github.com/jonase/eastwood/blob/7672782d748bcc15c0ea5bf1d913c3d47c9e0810/src/eastwood/lint.clj#L391-L396
* `lint-namespace` calls `lint-ns` https://github.com/jonase/eastwood/blob/7672782d748bcc15c0ea5bf1d913c3d47c9e0810/src/eastwood/lint.clj#L353
* `lint-ns` calls `analyze/analyze-ns` https://github.com/jonase/eastwood/blob/7672782d748bcc15c0ea5bf1d913c3d47c9e0810/src/eastwood/lint.clj#L162
* And accordingly the `analyze-ns` ns uses tools.analyzer: https://github.com/jonase/eastwood/blob/7672782d748bcc15c0ea5bf1d913c3d47c9e0810/src/eastwood/analyze_ns.clj#L13

> This change has been verified and there's a QA

QA and tests can only prove the absence of known flaws. Please accept the fact that tools.analyzer, just like `clojure.core/require` and `clojure.core/eval` need a non-trivial amount of care to be thread-safe.

Evaluating an arbitrary set of namespaces in parallel and no specific dependency order is the opposite of that.

thumbnail:
Thanks for bringing this to our attention. 

Because the issue you describe doesn't occur in several large internal projects I did some digging. It turns out that the change works fine in our testing because the codebase is reloaded beforehand, so no (new) code is evaluated by running Eastwood. Probably eliminating the issue.

Formatting stack is designed to work on reloaded code. So while this _shouldn't_ be an issue, I don't feel comfortable applying this change without a construct to ensure that has happened.

This means that using `:parallel?` should be opt-in, archivable with future work on configuration.

For the `:exclude-linters` change; without a way to 'replace' a part of the default-config, the consumer can no longer turn these linters on. This is too pending on future work on configuration.

The last change (removing extra keys in reporting) has little effect (unless debugging the linter). So it'll be postponed until more work is needed on this piece.

vemv:
> It turns out that the change works fine in our testing because the codebase is reloaded beforehand, so no (new) code is evaluated by running Eastwood.

Not sure of how you invoke f-s these days, but in my case I tend to use its `:in-background? true` option.

So a rapid succession of `(refresh :after format!)` calls can create a race condition.

This is unrelated to the :parallel? option though. Anyway if you are curious, the root cause is that t.n lacks a global lock, which is a flaw in itself no matter what. I'm working on a drop-in t.n replacement which features a lock, and can be honored from f-s: [wip](https://github.com/reducecombine/.lein/blob/cfe5e7e46cc106a6fe1fb41f117ca75203c01bbe/scripts/formatting_stack/linters/eastwood.clj#L33-L58)

> Probably eliminating the issue.

A race between t.n and Eastwood was not my original concern to begin with. An isolated Eastwood invocation with :parallel? is dangerous (or at least, _not cautious_) because clojure namespaces can define arbitrary contents, which aren't necessarily designed to be reloaded in parallel.

You might not experience issues for a year and still you could have them any given day. Also it is extremely hard to try guessing how this will behave in _arbitrary codebases_, i.e. using  just one or two organization's repos can be bit of a small sample size, leading to various biases.

---

At a deeper level, it's worth reiterating that f-s, as designed, defaults to agnosticism. Its very goal is eliminating futile discussions between developers with different sensibilities. There's no absolute truth (like a `:parallel?` value), and hopefully that will remain a first-class concept for us! The only limit is when a default stop being cautious in face of reasoning and evidence. (Which doesn't limit configurability, only the default value itself)

So, very obviously I don't have a problem if you set `:parallel? true` in your own configs, or maybe for handiness you follow the same model I did with https://github.com/nedap/formatting-stack/pull/149 (tldr: Java system prop, leave the default unchanged, explicitly exercise in CI for an indefinite time)

PR #174, opened 2021-02-23
clj-kondo improvements

thumbnail:
## Brief

1. Re-enables 'redundant' linters.
   - The formatters are run before the linters. So these shouldn't trigger. If they trigger they are missed by the formatter, and thus should be shown.
2. enable parallelisation
3. remove custom cache dir
   - Initially introduced to prevent 'locking' conflicts, clj-kondo currently [has a retry](https://github.com/clj-kondo/clj-kondo/blob/13c05a793ce667ef27de9a211129cd43aa4498d2/src/clj_kondo/impl/cache.clj#L49-L51) making this redundant.
   - Allows running `clj-kondo --parallel --lint "$(lein classpath)"` to populate cache as expected.
4. chunk analysing entire classpath
   - big classpaths can result in OOM. Chunking over them mitigates this (problem observed without parallelisation too)

## QA plan

1. postfix version project.clj with -dev
2. lein install
3. run `(format-and-lint-project!)` in projects to assert correct behaviour

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/linters/kondo.clj
+++ b/src/formatting_stack/linters/kondo.clj
@@ -18,15 +18,8 @@
                :private-call         off ;; undesired
                :unresolved-symbol    off ;; can give false positives
                :unused-symbol        off ;; can give false positives
-               :unused-private-var   off ;; can give false positives
-               :unresolved-var       off ;; already offered by clj
-               :consistent-alias     off ;; already offered by how-to-ns
-               :duplicate-require    off ;; already offered by clean-ns
-               :unused-import        off ;; already offered by clean-ns
-               :unused-namespace     off ;; already offered by clean-ns
-               :unused-referred-var  off ;; already offered by clean-ns
-               :unresolved-namespace off ;; already offered by clean-ns
-               }
+               :unused-private-var   off} ;; can give false positives
+
    :lint-as   '{nedap.speced.def/def-with-doc clojure.core/defonce
                 nedap.speced.def/defn         clojure.core/defn
                 nedap.speced.def/defprotocol  clojure.core/defprotocol
@@ -61,9 +54,9 @@
                                             (-> (re-find #"\.cljs$" f)
                                                 boolean))))]
     (->> [(kondo/run! {:lint   clj-files
-                       :config (deep-merge default-options clj-options kondo-clj-options)
+                       :config (deep-merge default-options clj-options kondo-clj-options)})
                        ;; :lang is unset here, so as not to particularly prefer .clj over .cljc
-                       })
+
           (kondo/run! {:lint   cljs-files
                        :config (deep-merge default-options kondo-cljs-options)
                        :lang   :cljs})]
--- a/src/formatting_stack/linters/kondo.clj
+++ b/src/formatting_stack/linters/kondo.clj
@@ -11,6 +11,7 @@
 
 (def default-options
   {:cache     true
+   :parallel  true
    :linters   {:cond-else            off ;; undesired
                :missing-docstring    off ;; undesired
                :unused-binding       off ;; undesired
--- a/src/formatting_stack/linters/kondo.clj
+++ b/src/formatting_stack/linters/kondo.clj
@@ -11,7 +11,6 @@
 
 (def default-options
   {:cache     true
-   :cache-dir formatting-stack.kondo-classpath-cache/cache-dir
    :linters   {:cond-else            off ;; undesired
                :missing-docstring    off ;; undesired
                :unused-binding       off ;; undesired
--- a/worker/formatting_stack/kondo_classpath_cache.clj
+++ b/worker/formatting_stack/kondo_classpath_cache.clj
@@ -10,5 +10,6 @@
   (future
     (let [files (-> (System/getProperty "java.class.path")
                     (string/split #"\:"))]
-      (kondo/run! {:lint      files
-                   :cache     true}))))
+      (->> (partition-all 100 files) ;; prevent OOM when classpath is large
+           (mapv #(kondo/run! {:lint      %
+                               :cache     true}))))))
--- a/worker/formatting_stack/kondo_classpath_cache.clj
+++ b/worker/formatting_stack/kondo_classpath_cache.clj
@@ -4,22 +4,11 @@
   Only needs to be created once, as the classpath never changes."
   (:require
    [clj-kondo.core :as kondo]
-   [clojure.string :as string])
-  (:import
-   (java.io File)))
-
-;; Use kondo's official default config dir, so that we don't bloat consumers' project layouts:
-(def cache-parent-dir ".clj-kondo")
-
-;; Don't use .clj-kondo directly since it can be accessed concurrently (e.g. f-s + a second Kondo from VS Code):
-(def cache-subdir "formatting-stack-cache")
-
-(def cache-dir (str cache-parent-dir File/separator cache-subdir))
+   [clojure.string :as string]))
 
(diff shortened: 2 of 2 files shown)


src/formatting_stack/linters/kondo.clj
```diff
@@ -11,22 +11,15 @@
 
 (def default-options
   {:cache     true
-   :cache-dir formatting-stack.kondo-classpath-cache/cache-dir
```
vemv, line 14: Please extract the :cache-dir changes to a separate PR, preferably to an issue first so it can be discussed carefully.

As a brief experience report, java `nio` locking is not guaranteed to work (as reflected in its own javadocs), which by coincidence I happen to have verified with the **exact** same code that clj-kondo uses.

I suspect it has to do with OS/filesystem/encryption combination.

Please as a measure of caution don't introduce changes known to be broken on certain OSes.

src/formatting_stack/linters/kondo.clj
```diff
@@ -11,22 +11,15 @@
...
+   :parallel  true
```
vemv, line 14: 👍

src/formatting_stack/linters/kondo.clj
```diff
@@ -11,22 +11,15 @@
...
    :linters   {:cond-else            off ;; undesired
                :missing-docstring    off ;; undesired
                :unused-binding       off ;; undesired
                :private-call         off ;; undesired
                :unresolved-symbol    off ;; can give false positives
                :unused-symbol        off ;; can give false positives
-               :unused-private-var   off ;; can give false positives
```
vemv, line 21: formatting-stack was designed with a specific philosophy, namely to be integrated with REPLs, and gathering runtime info from the Clojure compiler and built-time introspection system, and from precise `eval`-based tools such as Eastwood.

`The formatters are run before the linters` is not quite a complete assessment because some of these faults can't be possibly fixed by a formatter. Part of the reason why these kondo linters are disabled is because they are already covered precisely by either the Clojure compiler (e.g. `unresolved-var`) or Eastwood.

Another reason is the occasional inaccuracy of clj-kondo's approach (otherwise very valuable and which I do imitate from time to time). Inaccurate warnings hinder a repl-friendly experience that can also be integrated with CI checks.

I'd much appreciate these removals were proposed and discussed in a per-case basis in an issue.

thumbnail: > formatting-stack was designed with a specific philosophy, namely to be integrated with REPLs, and gathering runtime info from the Clojure compiler and built-time introspection system, and from precise eval-based tools such as Eastwood.

Eastwood is not guaranteed to run and thus the code isn't guaranteed to be loaded. Especially when configuration becomes more flexible it's plausible that users will opt-out of Eastwood in REPL-use because of the latency.

> some of these faults can't be possibly fixed by a formatter.

That's why I removed these _defaults_. I think if clj-kondo comes across these issues they should be reported. Apparently we couldn't fix them, or the formatters didn't run.

> Part of the reason why these kondo linters are disabled is because they are already covered precisely by either the Clojure compiler or Eastwood

The duplicated linters are still disabled, the clj-kondo warning seems friendlier than the Clojure reader exception. Happy to disable the `unresolved-var` linter if you feel strongly about it; but it doesn't seem like a big deal.

> Another reason is the occasional inaccuracy of clj-kondo [...]

`:unused-private-var` is actually unintended removal in this case, shouldn't have been removed. 

The intended change was just about the linters backed by formatters.

vemv: > Eastwood is not guaranteed to run and thus the code isn't guaranteed to be loaded.

Crafting a custom 'stack' should also involve a sensible set of defaults that work holistically (i.e. no duplicates, and no false negatives either).

f-s offers a number of stack which all work holistically very accurately.

When crafting a custom stack that e.g. omitted Eastwood, it would be certainly OK to use different clj-kondo defaults as you suggest. But worsening f-s defaults for a _half-configured stack_ seems misguided to me.

Also, please observe that f-s has a diverse user base and accordingly we shouldn't make breaking changes (and preferrably also avoid changes in philosophy - people have shown interest in f-s as designed with a specific philosophy - mutating it to something else might be confusing)

Remember that CI can be set up depending on f-s, hence my extra caution.

As a constructive suggestion, I'd like to point out that a refactored config system (e.g. https://github.com/nedap/formatting-stack/pull/152 or an alternative design) could easily make this problem more elegant and less delicate to tackle.

I would encourauge  you to keep exploring config mechanisms, and see how much simplification can we squeeze out of it.

A specific tricky/valuable part is making our clj-kondo linter detect that eastwood is present, and auto-shrink its own config accordingly (probably this is easier to fix at an upper layer, e.g. strategies).

worker/formatting_stack/kondo_classpath_cache.clj
```diff
@@ -4,22 +4,13 @@
...
+   [clojure.string :as string]))
 
 (def classpath-cache
   (future
     (let [files (-> (System/getProperty "java.class.path")
                     (string/split #"\:"))]
-      (-> (File. cache-parent-dir cache-subdir) .mkdirs)
-      (kondo/run! {:lint      files
-                   :cache-dir cache-dir}))))
+      (->> (partition-all 100 files) ;; prevent OOM when classpath is large
```
vemv, line 13: Nice find here btw. You might want to report it upstream - they can plausibly value it, and preferrably we wouldn't have to use APIs in a particularly careful manner.

I'd also be a bit afraid of the runtime impact of making N consecutive `kondo/run!` invocations (i.e. it's possibly slower). My biggest `(-> (System/getProperty "java.class.path") (string/split #"\:") count)` has 800+ items in it, implying 9 run! invocations.

thumbnail: I reported this in Clojurians yesterday, but no responses unfortunately. Ill create an issue soon to see what the maintainer thinks.

---


The 9 invocations seem like a lot, but I don't think it's more significant than 1 invocation at 9times the size.

In my limited testing I could only compare runtime between creating the cache using `clj-kondo` command directly and the partitioned runs (because OOM). The wall-to-wall times were pretty comparable (~40s in my case). Lmk your experiences though!

vemv: I responded over Clojurians with a possibly useful insight.

Perhaps you can repro this deterministically by intentionally setting  a low Xmx (say, `512m`) and an intentionally large input?

Maybe clj-kondo could ensure that its `:lint` input is processed in a lazy manner - that way, it would fit in memory at all times.

vemv: > The 9 invocations seem like a lot, but I don't think it's more significant than 1 invocation at 9times the size.

Would have to measure. I'd also be afraid of constructing  an analysis cache without having the "whole picture" (the entire classpath) at hand. Otherwise it's well plausible that analysis will have a partial view and clj-kondo will be unable to accurately resolve namespaces, vars, etc.

thumbnail:
The changes in this PR will be addressed at a later moment.

Issue #168, opened 2021-01-21
Parallelize `formatting-stack.strategies/namespaces-within-refresh-dirs-only`


vemv:
## Problem statement

`formatting-stack.strategies/namespaces-within-refresh-dirs-only` is file-heavy and therefore slow. It can increase the cost of `(files-from-strategies strategies)` for whole-project analysis from 20s to 60s.

## Proposal

Parallelize it - it's easy: `xs ->> (filter ...)` -> `xs ->> (partitioning-pmap ...) (filter identity)`

## Alternatives and comparison

* Maybe check out if less file ops can be performed
  * This surely would mean performing logic over strings denoting filenames, which can be error-prone (e.g. in monorepos)

PR #178, opened 2021-03-16
Parallelize `namespaces-within-refresh-dirs-only`

vemv:
## Brief

Fixes nedap/formatting-stack#168

## QA plan

It can be observed that the included test passes even when undoing the fix.

I also verified the performance improvement described in https://github.com/nedap/formatting-stack/issues/168, in a repl that loads a large codebase:

#### Before

image

#### After

image

## Author checklist

* [ ] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [x] I have reviewed the PR
  * [x] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [x] Performance
  * [x] Breaking API changes

### DIFF
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -19,6 +19,7 @@
    [formatting-stack.strategies.impl.git-status :as git-status]
    [formatting-stack.util :refer [read-ns-decl require-lock try-require]]
    [nedap.speced.def :as speced]
+   [nedap.utils.collections.eager :refer [partitioning-pmap]]
    [nedap.utils.spec.api :refer [check!]])
   (:import
    (java.io File)))
@@ -166,13 +167,15 @@
                                        ;; Temporary, until https://git.io/Jeaah is a thing:
                                        true)))) refresh-dirs)]}
   (->> files
-       (filter (speced/fn [^string? filename]
-                 (if-not (read-ns-decl filename)
-                   true
-                   (let [file (-> filename File.)]
-                     (->> refresh-dirs
-                          (some (fn [dir]
-                                  (impl/dir-contains? dir file))))))))))
+       (partitioning-pmap (speced/fn [^string? filename]
+                            (if-not (read-ns-decl filename)
+                              filename
+                              (let [file (-> filename File.)]
+                                (when (->> refresh-dirs
+                                           (some (fn [dir]
+                                                   (impl/dir-contains? dir file))))
+                                  filename)))))
+       (filter identity)))
 
 (defn refactor-nrepl-available? []
   (locking require-lock
--- a/test/integration/formatting_stack/strategies.clj
+++ b/test/integration/formatting_stack/strategies.clj
@@ -1,9 +1,12 @@
 (ns integration.formatting-stack.strategies
   (:require
    [clojure.java.shell :refer [sh]]
+   [clojure.set :as set]
    [clojure.string :as string]
    [clojure.test :refer [deftest is testing]]
-   [formatting-stack.strategies :as sut])
+   [formatting-stack.strategies :as sut]
+   [formatting-stack.util :refer [rcomp]]
+   [nedap.speced.def :as speced])
   (:import
    (java.io File)))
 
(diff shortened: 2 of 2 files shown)


test/integration/formatting_stack/strategies.clj
```diff
@@ -235,3 +238,18 @@
         (sh "rm" (str rename-destination))
         (assert (-> deletable-file .exists))
         (assert (not (-> rename-destination .exists)))))))
+
+(deftest namespaces-within-refresh-dirs-only
+  (speced/let [^{::speced/spec (rcomp count (partial < 100))}
```
thumbnail, line 243: Why is this assertion here? It'll cause an unrelated spec failure when formatting stack exceeds 100 tracked source files.

test/integration/formatting_stack/strategies.clj
```diff
@@ -235,3 +238,18 @@
...
+               all-files (sut/all-files :files [])
+               result (sut/namespaces-within-refresh-dirs-only :files all-files)]
+    (is (seq result)
+        "Returns non-empty results (since f-s itself has namespaces within `src`, `test`, etc)")
+
+    (is (< (count result)
+           (count all-files))
+        "Doesn't include files outside the refresh dirs")
```
thumbnail, line 251: Relying on count for this assertion seems error prone. Paired with the next `is` (which does the described assertion), it's fine.

src/formatting_stack/strategies.clj
```diff
@@ -166,13 +167,15 @@
...
-                                  (impl/dir-contains? dir file))))))))))
+       (partitioning-pmap (speced/fn [^string? filename]
+                            (if-not (read-ns-decl filename)
+                              filename
+                              (let [file (-> filename File.)]
+                                (when (->> refresh-dirs
+                                           (some (fn [dir]
+                                                   (impl/dir-contains? dir file))))
+                                  filename)))))
+       (filter identity)))
```
thumbnail, line 178: `(remove nil?)` or `(filter some?)` May be a bit more descriptive.

thumbnail (approved):
Thanks for this one 🚀

test/integration/formatting_stack/strategies.clj
vemv: In fact the opposite, read it again. Mistaking `<` happens to the best of us ;p

The point is that `all-files` should return a high number of files, it's an assumption/requirement of this test. If it returned say 0-10 files the output would mismatch the expectations, and we could be getting a false negative.

This assertion must be placed here (and not in `all-files` itself) because there's no generic number that will be true for all projects.

test/integration/formatting_stack/strategies.clj
vemv: Note that any set _x_ is a subset of itself, so a count assertion is essential.

src/formatting_stack/strategies.clj
vemv: `filter some?` assumes that the given coll is not a coll of booleans. That always can change over time.

Similarly `remove nil?` keeps `false` values, which creates a tight assumption about the given coll's type.

`filter identity` is unequivocal in signaling that it only cares about Clojure's _truthiness_ semantics, which is the programmer's intent here.

vemv (commented):
I appreciate the cautiousness, always good to verbalise some reasoning 👍

test/integration/formatting_stack/strategies.clj
thumbnail: The rcomp + partial combination caught me off guard I guess 😅. Esp since the order of the arguments matter a lot if used in this case.


Anyway; sgtm!

test/integration/formatting_stack/strategies.clj
thumbnail: For sure. I was mainly targeting the description in this case.

The assertion is only testing of the result is smaller now. The next assertion asserts that it's a subset. Which is👌🏼 when combined.

Issue #157, opened 2020-04-24
`git-diff-against-default-branch` accepts non-existing branches as an argument


vemv:
## Expected behavior

For the following input:

```clj
(git-diff-against-default-branch :target-branch "I-dont-exist")
```

...A precondition should fail informing nicely that the given `:target` branch doesn't exist.

## Actual behavior

`nil` is returned, suggesting that the branch exists but there's an empty diff.

## Environment info

4.2.1

vemv:
Currrently` :target-branch` accepts things like `c2e3259c`, `HEAD~2`, `v0.19.0` etc. I'd keep that supprt while still throwing on any non-existing object.

thumbnail:
Good suggestion! We could solve this by trying to parse the ref, if it doesn't resolved (or is ambiguous), throw.

PR #179, opened 2021-03-16
Add `git-ref-exists?` assertion

vemv:
## Brief

Fixes nedap/formatting-stack#157

## QA plan

* Add:

image

* Run:

```clj
(formatting-stack.branch-formatter/format-and-lint-branch!)
```

...The error will be reported 'humanly', and `13 exceptions occurred` will be reported (proving that the exceptions are handled like any other, i.e. failing a linter run for CI)

## Author checklist

* [ ] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [x] I have checked out this branch and reviewed it locally, running it
* [x] I have reviewed the PR
  * [x] Correctness
  * [x] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [x] Breaking API changes

### DIFF
--- a/src/formatting_stack/strategies/impl.clj
+++ b/src/formatting_stack/strategies/impl.clj
@@ -100,3 +100,10 @@
               (-> f .getCanonicalPath)))
        (some #{(-> file .getCanonicalPath)})
        (boolean)))
+
+(speced/defn ^boolean? git-ref-exists?
+  "Does `ref` denote an existing git branch, tag or commit sha?"
+  [^string? ref]
+  (-> (sh "git" "rev-parse" ref)
+      :exit
+      zero?))
--- a/test/integration/formatting_stack/strategies/impl.clj
+++ b/test/integration/formatting_stack/strategies/impl.clj
@@ -3,7 +3,8 @@
    [clojure.test :refer [are deftest is testing]]
    [formatting-stack.strategies.impl :as sut])
   (:import
-   (java.io File)))
+   (java.io File)
+   (java.util UUID)))
 
 (deftest dir-contains?
   (are [dirname filename expected] (= expected
@@ -33,3 +34,13 @@
     "src/../src/formatting_stack/strategies/impl.clj")
 
   (is (spec-assertion-thrown? ::sut/existing-files (sut/absolutize "git" ["I_dont_exist"]))))
+
+(deftest git-ref-exists?
+  (are [input expected] (testing input
+                          (is (= expected
+                                 (sut/git-ref-exists? input)))
+                          true)
+    "f374fcc"                  true
+    "master"                   true
+    "v4.3.0"                   true
+    (-> (UUID/randomUUID) str) false))
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -84,6 +84,8 @@
              ;; excluding Unmerged, Deleted, Type-changed, Broken (pair), and Unknown files
              impl          (impl/file-entries git-command "diff" "--name-only" "--diff-filter=ACMR" target-branch)
              blacklist     (git-not-completely-staged :files [])}}]
+  (assert (impl/git-ref-exists? target-branch)
+          (str (pr-str target-branch) " was not recognised as an existing git branch, tag or commit sha."))
   (->> impl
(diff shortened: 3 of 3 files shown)


src/formatting_stack/strategies/impl.clj
```diff
@@ -100,3 +100,10 @@
               (-> f .getCanonicalPath)))
        (some #{(-> file .getCanonicalPath)})
        (boolean)))
+
+(speced/defn ^boolean? git-ref-exists?
+  "Does `ref` denote an existing git branch, tag or commit sha?"
+  [^string? ref]
+  (-> (sh "git" "rev-parse" ref)
```
thumbnail, line 107: `#'formatting-stack.strategies/git-command` is a constant for `"git"` (albeit you can't actually use it here)

test/integration/formatting_stack/strategies/impl.clj
```diff
@@ -33,3 +34,13 @@
     "src/../src/formatting_stack/strategies/impl.clj")
 
   (is (spec-assertion-thrown? ::sut/existing-files (sut/absolutize "git" ["I_dont_exist"]))))
+
+(deftest git-ref-exists?
+  (are [input expected] (testing input
+                          (is (= expected
+                                 (sut/git-ref-exists? input)))
```
thumbnail, line 41: Is the `testing` necessary here? Using `is` inside `are` causes superfluous assertions. (sometimes _definitely_  worth it btw.)

Butreporting is fine without `testing`, because the input is already shown clearly;
```clj
(clojure.test/is ((constantly false) "My input"))

FAIL in () (form-init10638740795635241686.clj:1)
expected: ((constantly false) "My input")
  actual: false
=> false
```


wdyt?

thumbnail (approved):
💪  Thanks for this!

src/formatting_stack/strategies/impl.clj
vemv: Interesting. `git-command` has no big purpose, simple it's something DRYed out of habit.

FWIW there's still no need of making the command configurable (and maybe it can be better configured by simply manipulating the PATH)

test/integration/formatting_stack/strategies/impl.clj
vemv: I find it a good habit, given the following considerations:

* I prefer looking at the `testing`  strings vs. the "inlined code replacements" (for lack of a better name) that clojure.test performs
  * looking at the what (vs. the how) tends to suit better my style of thinking
* having a `testing` readily available is _designed for growth_
  * e.g. more args may be added tomorrow
    * if there wasn't a `testing` already there, one can be lazier about introducing a `testing` (and having a noisier diff later)
* `testing` forms part of the snippet I use and foster ([1](https://raw.githubusercontent.com/zenmacs/.emacs.d/fbed5d4b5e44ad65282c10a1b647d4fa014765e6/snippets/clojure-mode/are), [2](https://raw.githubusercontent.com/zenmacs/.emacs.d/fbed5d4b5e44ad65282c10a1b647d4fa014765e6/snippets/clojure-mode/dft))
  * `are` is admittledly a somewhat delicate construct.
  * by using a snippet, one can use `are` without having to think about intricacies
    * having to remember to add/remove `testing` breaks that 'abstraction'

Perhaps someday we'll have a higher-level `are` that is flawless, meanwhile I favor using it in an homogeneous manner that doesn't make people think too much :)

src/formatting_stack/strategies/impl.clj
thumbnail: Yes. I agreed. Might want to deprecate/remove that def at a later stage. But it's no issue rn

test/integration/formatting_stack/strategies/impl.clj
thumbnail: Your first and last considerations boil down to personal preferences. The second consideration is valid, but should be considered on a case to case basis. Given the triviality of this unit test I don't think it needs a more extensive `are`.

However; this part of the diff doesn't warrant a big discussion either.

Issue #158, opened 2020-04-25
f-s must be run binding *read-eval* to false


vemv:
## Problem statement

Some .clj files can contain the reader-eval syntax. https://clojure.org/guides/weird_characters#_reader_eval says:

> Note that #= is not an officially supported feature of the reader, so you shouldn’t rely on its presence in future versions of Clojure.

People can include this #= syntax in their .clj files, for reasons (personally I don't recommend so).

When tools.reader, used by our `strategies` (and potentially used by any of the libs we wrap) encounters a `#=()`, it will actually trigger Clojure evaluation, while `*ns*` is bound to a different ns to the one being read.

So if foo.clj contains `#=(release-sharks)`, whenever some other ns attempts to read this file, tools.reader will invoke `release-sharks`, attempting to resolve it to `*ns*` (not `foo`).

This is wrong at several levels:

* Running code is undesired to begin with
  * Can create very surprising behavior - e.g. running f-s should not cause application-level side-effects
* because `*ns*` is bound to a different ns to the one being read, it's very unlikely that function invocations will succeed
  * This causes an exception, making the whole file unreadable.

## Proposal

Bind `*read-eval*` to `false` in the right place, such that it's always false when running any strategy, linter, formatter, etc.

## AC

Ideally, a .clj file containing `#=(foo)` can be formatted and linted.

Currently, such a file would be excluded at a `strategies` level:

https://github.com/nedap/formatting-stack/blob/831ffdd5c4b674d2c1af3f18ad0088a3f92f3dcb/src/formatting_stack/strategies/impl.clj#L66

## Alternatives and comparison

* Bind `*ns*` to the ns being read
  * This complects code loading/running (a concern of e.g. tools.namespace) with formatting/linting. Generally f-s should not trigger application-level side-effects, perform excessive `require`s, etc.

vemv:
⚠️to be done after https://github.com/nedap/formatting-stack/pull/152

PR #180, opened 2021-03-17
Forbid read-eval syntax in formatted/linted files

vemv:
#### Forbid read-eval syntax in formatted/linted files

This particularly targets Eastwood (since it's the only member of our stack that evals code, and therefore can run read-eval syntax) and tools.reader (which we internally use for various purposes).

It also is bound at `formatting-stack.core` level in case any other mechanisms happen to be sensitive.

Fixes nedap/formatting-stack#158

## QA plan

* Lint formatting-stack itself via the `project` or `branch` formatter
  * Note that Eastwood will complain due to the two test namespaces that intentionally fail (functional.formatting-stack.linters.eastwood.examples.read-eval, functional.formatting-stack.linters.eastwood.examples.read-eval-2)
  *  Note that they must be placed in `test` and not `test-resources`, since ns parsing (and refresh-dirs parsing) is part of how our Eastwood integration works
    * This can be slightly inconvenient in a future when we dogood f-s in CI (not done atm), by then we could simply exclude these two files with a `strategy`.

## Author checklist

* [ ] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [ ] Documentation
    * Worth adding an explanation to the https://github.com/nedap/formatting-stack/wiki/FAQ
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes
    * Strictly speaking this would break a consumer that used `#=()` syntax. However `#=` is not part of the Clojure API: it's absent from https://clojure.org/reference/reader#_dispatch , https://clojure.org/guides/weird_characters
    * I do think it's good that f-s communicates that using `#=()` is a problem. My experience was that `#=()` can be detected and fixed.

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/formatters/clean_ns/impl.clj
+++ b/src/formatting_stack/formatters/clean_ns/impl.clj
@@ -10,7 +10,8 @@
    (clojure.lang Namespace)))
 
 (speced/defn safely-read-ns-contents [^string? buffer, ^Namespace ns-obj]
-  (binding [tools.reader/*alias-map* (ns-aliases ns-obj)]
+  (binding [tools.reader/*read-eval* false
+            tools.reader/*alias-map* (ns-aliases ns-obj)]
     (tools.reader/read-string {:read-cond :allow
                                :features  #{:clj}}
                               (str "[ " buffer " ]"))))
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -2,9 +2,12 @@
   (:require
    [clojure.java.io :as io]
    [clojure.string :as str]
+   [clojure.tools.reader :as tools.reader]
    [eastwood.lint]
+   [formatting-stack.formatters.clean-ns.impl :refer [safely-read-ns-contents]]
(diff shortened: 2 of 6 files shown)


vemv:
Not ready yet:

_defecord constructor syntax_ (#formatting_stack.strategies.foo.Foo[:a]) inherently uses read-eval. This alone shouldn't prevent a file from being formatted/linted.

Instead one could detect if there's `#=` in a given file (using e.g. rewrite-clj), and bind `*read-eval*` accoringly.

vemv:
Personally I'm not that interested in forbidding read-eval anymore. Or at least under a direct approach. It just gives too many problems IME - read-eval is sprinkled throughout Clojure internals so disabling it can be problematic.

Another approach would be linting this 'syntactically' i.e. with rewrite-clj.

Issue #167, opened 2020-12-09
improve git tests


thumbnail:
## Problem statement

https://github.com/nedap/formatting-stack/blob/6a82c0d402237ededfa5c7e95eee82e114b3346b/test/integration/formatting_stack/strategies.clj#L24-L28

This hinders running tests during development. It requires commiting files before running the tests which is unexpected and causes (specific) tests to be ignored until the very end. 

## Proposal

Create a tmp folder, which is filled with the needed files by a fixture. This'll ensure a clean sheet while it has no impact on the project's git status.

java has `File#deleteOnExit` which can be leveraged to cleanup even after failure

## Alternatives and comparison

Migrate to integration test, which'll run against known good output of the `git`-commands, without running them in the shell. Would also remove the need for `git` while running the tests.

vemv:
Agreed that the problem is pretty painful. The first proposal sgtm!

vemv:
I gave a quick shot to this and it seems to me that one would have to create a new project (`git init`) from scratch.

That would imply more setup/teardown steps, touching folders _outside the pwd_, etc so it might be excessively intrincate.

WDYT of wrapping the deftest in a `(when (System/getenv "CI") ...`?

thumbnail:
> That would imply more setup/teardown steps

Yes, but a simple `mkdir` + `git init` for setup, and `rm -rf` should suffice. 

> touching folders outside the pwd

This is not needed. Just make sure the folder is `.gitignore`. Locally i verified i could initialise a new git repo inside a ignored folder _inside_ the pwd :)

> WDYT of wrapping the deftest in a `(when (System/getenv "CI")` ...?

This is a fine alternative; but makes testing the git-related changes locally harder, and might throw off new contributors (in case of a passing `lein test` locally, but failing on CI for tests not even in your report.)


I'd prefer to explore the nested repo approach first, but am fine to settle with the pragmatic approach of disabling the tests locally if that doesn't work out.

vemv:
> Yes, but a simple mkdir + git init for setup, and rm -rf should suffice.

Will try sometime soon 👍

thumbnail:
I put a WIP for the suggestion above in [`167-improve-git-tests`](https://github.com/nedap/formatting-stack/tree/167-improve-git-tests). The code is not final in any way.

With that suggestion there's an issue with the `git-diff-against-default-branch`-test: It relies on 'a large, diverse corpus', which is not easy to replicate. 

[...]

PR #183, opened 2021-03-19
CI: perform `set-refresh-dirs`

vemv:
## Brief

Fixes the build broken in 87af50a8b6febdbcaa6c9908905c1e3c6024282e

Uses an approach outlined in the Lein template that this project uses.

#### Note

* Running the test suite from a repl/ide will pass, since a dev environment has the `refresh-dirs` set via dev.clj
* `lein test` or `lein with-profile -dev test` performed locally will keep failing, because `CI` is unset
  * That can be fixed by preceding the command with `CI=1`.
  * It also could be fixed in a manner similar to https://github.com/nedap/formatting-stack/issues/167#issuecomment-802047785

## QA plan

`CI=true lein with-profile -user,-dev do clean, test` should pass

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/dev/dev.clj
+++ b/dev/dev.clj
@@ -9,7 +9,8 @@
    [criterium.core :refer [quick-bench]]
    [lambdaisland.deep-diff]))
 
-;; the "worker" source-path must be excluded.
+;; * the "worker" source-path must be excluded.
+;; * if updating this, please check if `formatting-stack.global-test-setup` also needs updating.
 (set-refresh-dirs "src" "test" "dev")
 
 (defn prepare-tests []
--- a/test/formatting_stack/global_test_setup.clj
+++ b/test/formatting_stack/global_test_setup.clj
@@ -1,0 +1,15 @@
+(ns formatting-stack.global-test-setup
+  (:require
+   [clojure.tools.namespace.repl :refer [set-refresh-dirs]]))
+
+(when (System/getenv "CI")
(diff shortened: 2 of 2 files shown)


thumbnail:
> It also could be fixed in a manner similar to https://github.com/nedap/formatting-stack/issues/167#issuecomment-802047785

Why is it desirable not to fix this issue locally?

vemv:
> Why is it desirable not to fix this issue locally?

Because the proposed `(set-refresh-dirs "src" "test")` would easily conflict with that defined in dev.clj, which would lead to opaque issues

(Note that `locally` is a bit more broader than reality; running tests from the repl will work)

vemv:
...perhaps we can change it to

```clj
(when-not (seq refresh-dirs)
  (set-refresh-dirs "src" "test"))
```

I wouldn't like it 100%, but it's not fragile either and would fix `lein test`

thumbnail:
That sounds like a good idea!

vemv:
Thanks! Ready

image

Issue #139, opened 2020-03-02
Cover c2b3c4af072f16d8fdb63ee3ca561a9bcca86e46


vemv:
https://github.com/nedap/formatting-stack/pull/137/commits/c2b3c4af072f16d8fdb63ee3ca561a9bcca86e46 needs some coverage, although it's not trivial to manipulate the cwd under the JVM.

PR #184, opened 2021-03-20
Cover monorepo support with an e2e test

vemv:
## Brief

Fixes nedap/formatting-stack#139

Emulates a true monorepo via a subproject. i.e., uses formatting-stack's .git, but when running the subproject's test suite, its pwd will be more nested than the main suite's pwd.

## QA plan

* Remove `(cons "--full-name")` (as seen in the PR diff), commit that
* Note how the normal `lein test` passes, but `e2e.sh` doesn't.

Per my personal CircleCI account:

image

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/e2e/monorepo-support/test/monorepo_support/core_test.clj
+++ b/e2e/monorepo-support/test/monorepo_support/core_test.clj
@@ -1,0 +1,31 @@
+(ns monorepo-support.core-test
+  (:require
+   [clojure.set :as set]
+   [clojure.test :refer [are deftest is testing use-fixtures]]
+   [formatting-stack.strategies])
+  (:import
+   (java.io File)))
+
+(deftest e2e
+  (let [all (set (formatting-stack.strategies/all-files :files []))
+        staged (set (formatting-stack.strategies/git-completely-staged :files []))
+        unstaged (set (formatting-stack.strategies/git-not-completely-staged :files []))
+        staged+unstaged (into staged unstaged)]
+    (is (seq all))
+    (is (seq staged))
+    (is (seq unstaged))
+
+    (is (not= staged unstaged))
+    (is (> (count all)
+           (count staged+unstaged)))
+    (is (set/superset? all staged+unstaged))
+
+    (doseq [filename all
+            :let [file (File. filename)
+                  absolute (-> file .getAbsolutePath)]]
+      (is (= filename
+             absolute)
+          "Returns absolutized filenames, which is important for monorepo support")
+
+      (is (-> file .exists)
+          "Said absolutized filenames reflect files that actually exist"))))
--- a/src/formatting_stack/strategies.clj
+++ b/src/formatting_stack/strategies.clj
@@ -36,12 +36,16 @@
                        (map git-status/remove-deletion-markers)
                        (impl/absolutize git-command)
                        (set))
-          tracked (->> (impl/file-entries git-command "ls-files" "--full-name")
-                       (impl/absolutize git-command)
-                       (remove deleted))
-          untracked (->> (impl/file-entries git-command "ls-files" "--full-name" "--others" "--exclude-standard")
-                         (impl/absolutize git-command)
-                         (remove deleted))]
+          ;; DRY the use of the `--full-name` option which is especially important:
+          ls-files (speced/fn [^list? args]
+                     (->> args
+                          (cons "--full-name")
+                          (cons "ls-files")
+                          (apply impl/file-entries git-command)
+                          (impl/absolutize git-command)
+                          (remove deleted)))
+          tracked (ls-files ())
+          untracked (ls-files '("--others" "--exclude-standard"))]
       ;; Second `binding`, to ensure correct results
       (binding [impl/*skip-existing-files-check?* false]
         (speced/let [^::impl/existing-files corpus (into tracked untracked)]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,8 @@ jobs:
       - setup-env
       - run:
           command: << parameters.lein_test_command >>
+      - run:
+          command: .circleci/e2e.sh
   deploy:
     executor: openjdk8
     steps:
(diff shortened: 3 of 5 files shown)


src/formatting_stack/strategies.clj
```diff
@@ -36,12 +36,16 @@
...
+          ;; DRY the use of the `--full-name` option which is especially important:
+          ls-files (speced/fn [^list? args]
+                     (->> args
+                          (cons "--full-name")
+                          (cons "ls-files")
+                          (apply impl/file-entries git-command)
+                          (impl/absolutize git-command)
+                          (remove deleted)))
+          tracked (ls-files ())
+          untracked (ls-files '("--others" "--exclude-standard"))]
```
thumbnail, line 48: using `& args` would result in a `list?` for `args` too, and make the api more elegant; 

i.e.
```clj
(ls-files)
(ls-files "--others" "--exclude")
```

e2e/monorepo-support/test/monorepo_support/core_test.clj
```diff
@@ -0,0 +1,30 @@
+(ns monorepo-support.core-test
+  (:require
+   [clojure.set :as set]
+   [clojure.test :refer [are deftest is testing use-fixtures]]
+   [formatting-stack.strategies])
+  (:import (java.io File)))
```
thumbnail, line 6:
```suggestion
  (:import 
   (java.io File)))
```

thumbnail (approved):
⚡  Thanks for this one, 1 small suggestion; other than that LGTM

src/formatting_stack/strategies.clj
vemv: It's a good suggestion, however two considerations apply:

* I prefer that `list?` is part of the explicit contract, and not something that can go away unintentionally (and perhaps break due to vector<->list differences)
* `& args` implies `apply` will be used internally, which performs worse

Performance tends to be a good 'decider' in case of doubt. I also tend to favor performant code regardless of whether is needed, simply as a habit that can pay off in unsuspected places.

vemv:
Rebased and formatted

src/formatting_stack/strategies.clj
thumbnail: > I prefer that list? is part of the explicit contract, and not something that can go away unintentionally (and perhaps break due to vector<->list differences)

You can still spec `args`, explicitly:
```clj
(speced/fn [& ^list? args] ...)
```

> & args implies apply will be used internally, which performs worse

This fn is not run in a tight loop or anything, so it's no need for a premature optimalisation.

vemv: > This fn is not run in a tight loop or anything, so it's no need for a premature optimalisation.

Over the years I've found a no-broken-windows approach to performance is cheap and effective enough: if performance can be gained by applying simple habits, then it should.

The fixed cost of `(apply f)` is 2x as slow as that of `(f)`. For an input large enough (here, for example: a large monorepo), that 2x often becomes noticeable.

Anyway the point I'm defending is precisely the opposite of fine-grained reasoning: in case of doubt (especially given two equally defensible choices), one can use an _objective measurement_ as a decider (and guide for everyday programming).

---

It's a similar reasoning I use to design speced.def (with spec->type inference), pay attention to compiler/JVM flags in our libs, etc. All those can be seen as 'irrelevant' but they do add up and free one of having to chase false leads when a real problem arises.

thumbnail: > Over the years I've found a no-broken-windows approach to performance is cheap and effective enough: if performance can be gained by applying simple habits, then it should.

If _tangible_ performance can be gained [...] Sure. In this case it's sub nanosecond. Which, in a higher order language, is not tangible whatsoever.

> For an input large enough

The api in your case has 0..2 arguments, a large monorepo affect that. Keep in mind that using `apply` is not needed at all.

vemv: Please don't get past of what the no-broken-windows principle stands for: one follows habits, no matter what, no exceptions, because breaking them opens a slippery slope. It's that slippery-slope what I'm afraid of: by setting concrete precedents, future changes to PRs can similarly disregard performance. Individual reasonings from the past can stop holding true, in face of code changes, concurrent PRs, a diversity of authors, unforeseen use cases, etc.

So please let's not break that window?

And also, please observe that the alternative offers no tangible benefits. One is erasing a ` ()` and adding a ` &`. Performance worsens and nothing else changes. I'm not really _against_ your proposal, but we should be able to have an objective decider when in doubt.

Let's take this constructively as an opportunity to reflect what a code review is _for_? It rarely is for altering a diff, character by character, until it looks exactly as a reviewer would have authored himself.

Interestingly, the long-term (and original) goal of formatting-stack is precisely elliminating these churn-y code reviews: if a PR passes sufficiently smart checks (including test coverage analysis, which to be fair I have barely leveraged) and satisfies business requirements, it should be essentially fine.

> Keep in mind that using apply is not needed at all.

`&` implies `apply` at the compiler level, it's what is being invoked under the hood and why any defn with `&` is slower.

While one can reason whether some optimization would be irrelevant, one can also observe that _fine-grained reasoning doesn't scale_, which is another key driver for formatting-stack (and speced.def). By using heavy linting, assertions, etc one can catch things that otherwise an 'outdated reasoning' would have not caught.

thumbnail: It is clear that this discussion is not going anywhere. I do not share your personal views expressed here, but given the impact of the change (better said; the lack thereof), I'll leave this be.

PR #185, opened 2021-03-22
Perform git-related tests in different repo

thumbnail:
## Brief

Suggestion as discussed in #167.
 

## QA plan

1. run tests
2. change some files
3. run tests again

Note no test failure related to git-status not being pristine

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 /.lein-*
 /checkouts
 /classes
+/git-integration-testing
 /logs/
 /target
 build/
--- a/test/integration/formatting_stack/strategies.clj
+++ b/test/integration/formatting_stack/strategies.clj
@@ -1,21 +1,22 @@
 (ns integration.formatting-stack.strategies
   (:require
+   [clojure.java.io :as io]
    [clojure.java.shell :refer [sh]]
    [clojure.set :as set]
    [clojure.string :as string]
    [clojure.test :refer [deftest is testing use-fixtures]]
    [formatting-stack.strategies :as sut]
-   [formatting-stack.test-helpers :as test-helpers]
+   [formatting-stack.test-helpers :as test-helpers :refer [git-integration-dir]]
    [formatting-stack.util :refer [rcomp]]
    [nedap.speced.def :as speced])
   (:import
    (java.io File)))
 
-(def ^File deletable-file (File. "git-integration-testing/test-resources/deletable.clj"))
+(def ^File deletable-file (io/file git-integration-dir "test-resources" "deletable.clj"))
 
-(def ^File rename-destination (File. "git-integration-testing/test-resources/deletable_moved.clj"))
+(def ^File rename-destination (io/file git-integration-dir "test-resources" "deletable_moved.clj"))
 
-(def ^File creatable-filename (File. "git-integration-testing/test-resources/createable.clj"))
+(def ^File creatable-filename (io/file git-integration-dir "test-resources" "creatable.clj"))
 
 (def creatable-contents (pr-str '(ns foo)))
 
@@ -54,7 +55,7 @@
   (let [n (count corpus)]
     (is (pos? n))
     (doseq [^String filename corpus
-            :let [f (File. filename)]]
+            :let [f (io/file filename)]]
       (is (-> f .exists))
       (testing "It emits absolutized filenames"
         (is (= filename
@@ -242,4 +243,4 @@
 
     (is (set/subset? (set result)
                      (set all-files))
-        "Is a substractive (and not additive) mechanism")))
+        "Is a subtractive (and not additive) mechanism")))
--- a/test/integration/formatting_stack/strategies.clj
+++ b/test/integration/formatting_stack/strategies.clj
@@ -3,26 +3,46 @@
    [clojure.java.shell :refer [sh]]
    [clojure.set :as set]
    [clojure.string :as string]
-   [clojure.test :refer [deftest is testing]]
+   [clojure.test :refer [deftest is testing use-fixtures]]
    [formatting-stack.strategies :as sut]
+   [formatting-stack.test-helpers :as test-helpers]
    [formatting-stack.util :refer [rcomp]]
    [nedap.speced.def :as speced])
   (:import
    (java.io File)))
 
-(def ^File deletable-file (File. "test-resources/deletable.clj"))
+(def ^File deletable-file (File. "git-integration-testing/test-resources/deletable.clj"))
 
-(def ^File rename-destination (File. "test-resources/deletable_moved.clj"))
+(def ^File rename-destination (File. "git-integration-testing/test-resources/deletable_moved.clj"))
 
-(def ^String createable-filename "test-resources/createable.clj")
+(def ^File creatable-filename (File. "git-integration-testing/test-resources/createable.clj"))
 
 (def creatable-contents (pr-str '(ns foo)))
 
-(assert (-> deletable-file .exists))
+(defn cleanup-testing-repo! []
+  (-> rename-destination .delete)
+  (-> creatable-filename .delete)
 
-(assert (not (-> rename-destination .exists)))
+  (sh "git" "reset" "--" "test-resources/deletable.clj")
+  (sh "git" "reset" "--" "test-resources/deletable_moved.clj")
+  (sh "git" "reset" "--" "test-resources/creatable.clj")
+  (sh "git" "checkout" "test-resources/deletable.clj")
 
-(assert (not (-> createable-filename File. .exists)))
+  (assert (-> deletable-file .exists))
+  (assert (not (-> rename-destination .exists)))
+  (assert (not (-> creatable-filename .exists))))
+
+(use-fixtures :once
+  test-helpers/with-git-repo
+
+  (fn [t]
+    (assert (-> deletable-file .exists))
+
+    (assert (not (-> rename-destination .exists)))
+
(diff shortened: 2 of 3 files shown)


test/integration/formatting_stack/strategies.clj
```diff
@@ -49,7 +61,7 @@
       (-> deletable-file .delete)
       (expect-sane-output! (sut/all-files :files []))
       (finally
-        (sh "git" "checkout" (str deletable-file))
+        (sh "git" "checkout" ".")
```
thumbnail, line 64: Other tests already relied on `.`, since we run these commands in a copied (temporary) repo **and** assert `git pristine` before, i think this is fine.

src/formatting_stack/strategies/impl.clj
```diff
@@ -27,7 +31,9 @@
 
 (speced/defn ^::file-entries file-entries
   [& args]
-  (->> args (apply sh) :out string/split-lines (filter seq)))
+  ;;; xxx
+  (let [r (apply sh (conj (vec args) :dir *pwd*))]
+    (->> r :out string/split-lines (filter seq))))
```
thumbnail, line 36: There should be a better way to add the `:dir` option here.

test/integration/formatting_stack/strategies.clj
vemv: I don't know, it would be very frustrating if some programming error meant that one's WIP was erased.

vemv (commented):
The approach seems sound to me, but I'd make sure that no WIP is possibly lost

test/integration/formatting_stack/strategies.clj
thumbnail: True, but I don't think it's likely. However; simply introducing another constant here should be doable.

.gitignore
```diff
@@ -28,3 +28,4 @@ pom.xml
 pom.xml.asc
 profiles.clj
 repl-port
+/tmp
```
vemv, line 31: maybe we can use `git-integration-testing`, so that the intent is clearer and don't end up storing unrelated stuff under `tmp`

test/integration/formatting_stack/strategies.clj
```diff
@@ -1,28 +1,53 @@
...
 
-(def ^String createable-filename "test-resources/createable.clj")
+(def ^File rename-destination (File. "tmp/test-repo/test-resources/deletable_moved.clj"))
+
+(def ^String creatable-filename "tmp/test-repo/test-resources/creatable.clj")
 
 (def creatable-contents (pr-str '(ns foo)))
 
-(assert (-> deletable-file .exists))
+(defn cleanup-repo []
```
vemv, line 27: maybe can be named `cleanup-testing-repo!` so that it's clearer that the main repo is not intended to be cleaned up

src/formatting_stack/strategies/impl.clj
thumbnail: I implemented a rudimentary approach where I use 1 impl for the `git`-command here; 15ba8e1

That suggestion also removes the `git-command`-def, and makes sure any git-command is executed in the right pwd. It removes /some/ flexibility from `impl/absolutize`, but we can extend/restore that if we ever need it in the future.

WDYT @vemv ?

src/formatting_stack/strategies/impl.clj
```diff
@@ -38,13 +48,12 @@
...
                                              (-> s File. .exists)))))
 
-(speced/defn ^::existing-files absolutize [command, ^::file-entries file-entries]
-  (let [toplevel-fragments (case command
-                             "git" (-> (sh "git" "rev-parse" "--show-toplevel")
-                                       (:out)
-                                       (string/split #"\n")
-                                       (first)
-                                       (string/split separator-pattern)))]
+(speced/defn ^::existing-files absolutize [^::file-entries file-entries]
```
vemv, line 51: `strategies/impl` is git-agnostic (`"git"` is only ever handled over a `case`), while `stategies.clj` makes git part of its explicit api (e.g. a given defn is called `git-completely-staged`, not `completely-staged`).

So as of `master`, we're very clear about what we promise / intend to offer: an opt-in git interface, and some agnostic implementation details.

I wouldn't deviate from that as f-s, should be usable in non-git environments (e.g. some enterprises).

---

(speaking of, we can make that `case` a multimethod instead - only an _open_ set is truly agnostic)

thumbnail: As eluded to here: https://github.com/nedap/formatting-stack/pull/185#discussion_r601425939; This change might look like flexibility is removed, but the flexibility is a false promise. The function is not open for extension, and only requires any consumer to provide a static argument (`"git"` in this case). Consumers without git would need their own replacements for the strategies anyway. 

This is an internal function anyway; which is subject to not public by definition. While I agree with the sentiment it's a opt-in git interface, i don't agree that this change affects that in any way.

vemv: >  but the flexibility is a false promise. 

It trivially becomes a true promise by refactoring a single `case` to a multimethod. We can be thankful for having discovered that via this PR.

> While I agree with the sentiment it's a opt-in git interface, i don't agree that this change affects that in any way.

It does increase the cost of implementing non-git strategies. As of this branch, if tomorrow someone created a PR with `hg` support, we'd plausibly have a hard time adapting internals, tests, etc because everything was more coupled to `git` than before.

thumbnail: > It trivially becomes a true promise by refactoring a single case to a multimethod. We can be thankful for having discovered that via this PR.

I don't think it's something we should pursue. The current naming of te strategies are not abstracted properly from git. Mecurial for example, doesn't have any notion of 'staged' files. Refactoring the case to a multimethod does not fix that.

> It does increase the cost of implementing non-git strategies.

I'm not sure about this. I prefer to have small, specific namespaces. If another vcs would be supported i would not put it in `formatting-stack.strategies`, but in a more specific one. Especially if you keep the opt-in promise in mind. 

I don't see how the amount of work would be affected.

vemv: Ok but let's rename absolutize -> git-absolutize

Also https://github.com/nedap/formatting-stack/pull/184 should be merged before this one, since that one precisely adds coverage for this sort of things

thumbnail: This change was mostly driven from the necessity to run the `git` command in a specific directory for both the test-suite and the code. Which in turn, presented the opportunity remove the `"git"`-definition and dependent code. This is extracted into 
 e61c276794575714f3015c489f8b8e241f8b204


I discovered `clojure.java.shell/with-sh-dir`, which perfectly suits our usecase here, so i undid related changes.

thumbnail:
@neominik I implemented your `git clone` suggestion, which worked like a charm. If you feel like a review, go for it !

test/integration/formatting_stack/strategies.clj
```diff
@@ -3,26 +3,46 @@
...
+   [clojure.test :refer [deftest is testing use-fixtures]]
    [formatting-stack.strategies :as sut]
+   [formatting-stack.test-helpers :as test-helpers]
    [formatting-stack.util :refer [rcomp]]
    [nedap.speced.def :as speced])
   (:import
    (java.io File)))
 
-(def ^File deletable-file (File. "test-resources/deletable.clj"))
+(def ^File deletable-file (File. "git-integration-testing/test-resources/deletable.clj"))
```
vemv, line 14: You can use `io/file` to avoid having `/`s around

test/integration/formatting_stack/strategies.clj
```diff
@@ -3,26 +3,46 @@
...
 
-(def ^File rename-destination (File. "test-resources/deletable_moved.clj"))
+(def ^File rename-destination (File. "git-integration-testing/test-resources/deletable_moved.clj"))
```
vemv, line 16: Would you use `"git-integration-testing"` as an extracted constant?

test/integration/formatting_stack/strategies.clj
```diff
@@ -3,26 +3,46 @@
...
+  (fn [t]
+    (assert (-> deletable-file .exists))
+
+    (assert (not (-> rename-destination .exists)))
+
+    (assert (not (-> creatable-filename .exists)))
+
+    (t)))
 
 (defn assert-pristine-git-status! []
```
vemv, line 47: Is `assert-pristine-git-status!` still useful?

(Might be part of the QA plan)

vemv (approved):
LGTM!

test/integration/formatting_stack/strategies.clj
thumbnail: While developing this PR did catch some nasty mistakes, so i'm happy to keep it in tbh.

vemv: SGTM, I just wasn't sure (that is, without running it) whether it was still a runing a valid assertion

josefigueroa-nedap (approved):
Tested and monitored during execution. It works pretty well! 👍

Issue #169, opened 2021-02-19
Idea: don't re-analyze files depending on their SHA


vemv:
## Problem statement

Any given linter (particularly Eastwood) can be slow.

For the use case of "running `branch-formatter` integrated with the repl", one can observe that one might be linting again files that haven't been modified.

For example:

* I'm working in a long-lived branch that has touched 30 files
  * accordingly, the `branch-formatter` strategy will dictate that 30 files must be analyzed per `lint!` invocation
* But I'm only actively working in a smaller subset, like 1 or 2 files
  * The other files touched by the branch may have been already analyzed, which makes them re-analyzing them redundant.

## Proposal

Implement a strategy that removes files if all these conditions are met:

* the given file has already been analyzed
* the SHA of the file's contents hasn't changed
* no linter warnings have been emitted for this particular file
  * this is important - otherwise one would show linter warnings just once, and then elide them in subsequent runs because the file hadn't changed.

## Caveats

It's plausible that fixing a given file's linting result does not depend entirely on the file itself, fixing it might be accomplished by modifying a _different_ file.

That could cause stale caches or such.

However I'm not immediately aware if such a use case. Still worth a think, on a per-linter basis.

## Alternatives and comparison

Unknown

cc/ @thumbnail

vemv:
@thumbnail I'm tempted to give a shot to the alternative you suggested. I have the vague feeling that it can be something thin to implement.

IOW, this wouldn't conflict with architectural redesigns - we could always scrap my work, but keep the integration tests.

Assuming the resulting PR is indeed thin, does this plan SGTY?

thumbnail:
I think a wrapper is the most viable option. Simply using lastModified as a cache key to prevent files from being linted, and merge previous/cached reports in the end result.

I think the branch/project formatter shouldn't use the cached versions by default, that way CI doesn't either.

PR #187, opened 2021-03-30
Eastwood: always use absolute filenames, internally

vemv:
## Brief

Part of https://github.com/nedap/formatting-stack/issues/169

As a prerequisite for #169, I found out that filenames coming from Eastwood reports should be _absolute_, otherwise they wouldn't match with filenames computed by a 'linter caching' mechanism. Of course, a key mismatch means no cache hits.

This is the data path as of `master`:

* f-s sends an absolutized filename (via our git strategies) to Eastwood
* Eastwood may return wither a File (which will be stringed, absolutized) or a string (which will be a relative filename)

This PR proposes making the data path always `absolutized filename` -> `absolutized filename`.

## QA plan

Verify that Eastwood keeps emitting reports when a fault arises, in vanilla repos and monorepos alike.

⚠️The report should be rendered with relativized filenames (as it's a friendly/concise UI), as always.

## Author checklist

* [ ] The build passes
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/linters/eastwood.clj
+++ b/src/formatting_stack/linters/eastwood.clj
@@ -7,6 +7,7 @@
    [formatting-stack.protocols.linter :as linter]
    [formatting-stack.util :refer [ns-name-from-filename]]
    [medley.core :refer [assoc-some deep-merge]]
+   [nedap.speced.def :as speced]
    [nedap.utils.modular.api :refer [implement]])
   (:import
    (java.io File)))
@@ -16,7 +17,7 @@
   (let [linters (remove #{:suspicious-test :unused-ret-vals :constant-test :wrong-tag}
                         eastwood.lint/default-linters)]
     (-> eastwood.lint/default-opts
-        (assoc :linters linters
+        (assoc :linters             linters
                :rethrow-exceptions? true))))
 
 (def parallelize-linters? (System/getProperty "formatting-stack.eastwood.parallelize-linters"))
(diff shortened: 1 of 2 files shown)


e2e/monorepo-support/test/monorepo_support/core_test.clj
```diff
@@ -22,7 +22,7 @@
 
     (doseq [filename all
             :let [file (File. filename)
-                  absolute (-> file .getAbsolutePath)]]
+                  absolute (-> file .getCanonicalPath)]]
```
vemv, line 25: Settling in using `getCanonicalPath` only seems a good idea:

* one less thing to remember
* importantly, it resolves symlinks, resulting in cache keys that can hit more often

From memory, libs like tools.namespace favor .getCanonicalPath

src/formatting_stack/linters/eastwood.clj
```diff
@@ -28,6 +32,14 @@
 (note that this prefix must not be passed to Eastwood itself).")
 
 (defn lint! [{:keys [options]} filenames]
+  {:post [(do
```
vemv, line 35: A glorified :such-that ;p

An alternative (or complementary) solution would be strengthening `::protocols.spec/filename` from `present-string?` to a string that denotes an absolutized, existing filename.

IIRC a very similar idea was rejected ~1y ago (can't find the link, sorry) so I'm fine with leaving things as-is

vemv (commented):
(Ready-ish but I should QA it)

src/formatting_stack/linters/eastwood.clj
thumbnail: > from present-string? to a string that denotes an absolutized, existing filename. IIRC a very similar idea was rejected ~1y ago 

I guess the reasoning would be something akin to "don't have side-effecting specs". I.e. a absolute filepath is fine (i.e. leading `/`), going through the filesystem is not.

It makes generating examples harder and makes a spec unpure / dependent on state outside of it's input

vemv:
(This branch has worked well for me in my personal setup. Will undust this PR soon enough)

vemv:
Rebased, gonna give it some extra QAing after this updating

Issue #132, opened 2020-02-24
report blacklist


thumbnail:
## Context

Sometimes a lint-warning won't need fixing (because edge cases, dev preference, etc.) 
We want to have a way to ignore recurring warnings from bothering the user.

the CI integration (#88) would need this, so the build does not fail and the warning is considered unimportant.

## Task

 1. create a extensible blacklist
 1. ignore reports matching the blacklist

## possible approaches
 1. a simple edn list with the `:line`, `:column`, `:filename`, and `:source` of all (to be) ignored reports
   - **con:** if the location of the warning changes, the ignore list has to be updated
   - we can automatically check whether all ignores are still triggered.
 1. metadata, by tagging a form with `^:formatting-stack.<linter>/ignored`.
   - moving the form doesn't break whitelist
   - self describing. it's right in the source.
   - **con:** clutters code with metadata irrelevant for the behavior of the program itself.
   - **con:** implementation might be more complex, as the violated form is not available in the reporter right now. Not all linters parse the source
  - **con:** not all reports address one specific form. e.g. `loc-per-file`, and `one-resource-per-ns`

vemv:
If the CI integration was analog to https://github.com/DeLaGuardo/clojure-lint-action , namely applying a diff over a PR's changes only, maybe https://github.com/nedap/formatting-stack/issues/132 wouldn't be that necessary.

As of lately, I'm thinking that a big CI step that analyzes the whole project is bound to be a nuisance to people, as always unrelated stuff will show up.

Requiring people to configure f-s seems not particularly attractive either.

(We can refine a https://github.com/DeLaGuardo/clojure-lint-action -like idea in a separate issue)

thumbnail:
> If the CI integration was analog to https://github.com/DeLaGuardo/clojure-lint-action , namely applying a diff over a PR's changes only, maybe #132 wouldn't be that necessary.

This issue overarches the CI issue because also (new) devs shouldn't be bothered by acknowledged / allowed warnings anyway. The main goal of this issue is to address unnecessary noise during development on (big / legacy) projects.

>  'm thinking that a big CI step that analyzes the whole project is bound to be a nuisance to people, as always unrelated stuff will show up.

I don't think so. If we can create a good blacklist which doesn't require a lot of maintenance etc., we can prevent this issue altogether, and keep the CI integration simple (exit with `(+ (count warnings))`)

> Requiring people to configure f-s seems not particularly attractive either.

[...]

PR #188, opened 2021-09-03
Upgrade Eastwood

vemv:
## Brief

Long due one!

Some of the changes in Eastwood that are most relevant to f-s:

* New reflection linter that makes our own linter redundant
* The linters that we used to disable have been thoroughly fixed and are not expected to have false positives
* Increased performance and reliability for REPL invocations
* New ways of silencing faults
  * https://github.com/jonase/eastwood#ignored-faults
  * https://github.com/jonase/eastwood#ignoring-linter-sub-kinds

## QA plan

For the Eastwood changes, I'd recommend linting large projects and verifying/fixing that the newly enabled linters aren't being problematic. Please report any false positives.

For the `runner` change, one can `(formatting-stack.core/format)` and observe that the background runner keeps doing its job.

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes
    * The `runner` change is "breaking" (changed a type from Future to Thread), at the same that ns is quite clearly not an API
    * e.g. we've never received a GH issue about the runner, so presumably nobody is hacking `runner`

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ commands:
             - v2-dependencies-{{ checksum "project.clj" }}
             # fallback to using the latest cache if no exact match is found
             - v2-dependencies-
-      - run: lein with-profile -dev,+ci,+test,+refactor-nrepl deps
+      - run: lein with-profile -user,-dev,+ci,+test,+refactor-nrepl deps
```
vemv, line 12: I added `-user` everywhere, lately I've realised that this makes commands more reproducible. It saves time that people can copy-paste command into their terminal, resting assured that ~/.lein won't interfere.

We use this in e.g. all the important https://github.com/clojure-emacs projects.

.circleci/config.yml
```diff
@@ -70,6 +70,8 @@ jobs:
           command: << parameters.lein_test_command >>
       - run:
           command: .circleci/e2e.sh
+      - run:
+          command: lein with-profile -user,+test eastwood
```
vemv, line 74: Good to self-lint. It's not very formatting-stacky to lint all files, but it's a good idea (self-reference tends to be logically weak)

project.clj
```diff
@@ -131,14 +134,12 @@
                                      ;; (and shipped with this refactor-nrepl):
                                      :plugins      [[cider/cider-nrepl "0.24.0"]]}
 
-             :parallel-eastwood     {:jvm-opts ["-Dformatting-stack.eastwood.parallelize-linters=true"]}
```
vemv, line 134: Eastwood already parallelizes linters now

(f-s' own testing aided that decision!)

src/formatting_stack/linters/eastwood/impl.clj
```diff
@@ -5,45 +5,6 @@
    [formatting-stack.protocols.spec :as protocols.spec]
    [nedap.speced.def :as speced]))
 
-(speced/defn ^boolean? contains-dynamic-assertions?
```
vemv, line 8: This fix made its way into Eastwood ~1y ago

worker/formatting_stack/background.clj
```diff
@@ -4,15 +4,23 @@
...
+                    (job)
+                    (catch Exception e
+                      (-> e .printStackTrace))
+                    (catch AssertionError e
+                      (-> e .printStackTrace))))
+                (Thread/sleep 50))))]
+    (-> f
+        Thread.
+        ;; Important - daemonize this thread, otherwise under certain conditions it can prevent the JVM from exiting.
+        ;; (We exercise this implicitly via `lein eastwood` in CI)
```
vemv, line 24: As the comment hints, the thing being fixed here is that a non-daemonized thread can leave the JVM hanging forever.

I certainly have experieced that with f-s a bunch of times, it's very annoying.

worker/formatting_stack/kondo_classpath_cache.clj
```diff
@@ -24,10 +24,10 @@
   (apply list runner-choice body))
 
 ;; Use kondo's official default config dir, so that we don't bloat consumers' project layouts:
-(def cache-parent-dir ".clj-kondo")
+(def ^String cache-parent-dir ".clj-kondo")
```
vemv, line 27: Fixes found by Eastwood

test/functional/formatting_stack/linters/eastwood.clj
```diff
@@ -27,11 +27,11 @@
 
       "test-resources/eastwood_warning.clj"
       (matchers/in-any-order
-       [{:source              :eastwood/warn-on-reflection
```
vemv, line 30: Maybe the linter rename is a breaking change 

OTOH any Eastwood upgrade can bring new warning types (e.g. `:eastwood/boxed-math`) so it to be expected that types can appear/disappear over time

.circleci/config.yml
thumbnail: This is a good idea, for all my 'reproducible' commands (`lein test`-aliases and such) I include `-user` too :)

test/functional/formatting_stack/linters/eastwood.clj
thumbnail: Well, it would if we had silencing / blacklisting implemented (see #132). So I think this'll be fine :)

src/formatting_stack/linters/eastwood.clj
```diff
@@ -53,8 +45,7 @@
                             :filename            (if (string? uri-or-file-name)
                                                    uri-or-file-name
                                                    (-> ^File uri-or-file-name .getCanonicalPath)))))
-         (concat (impl/warnings->reports output)
```
thumbnail, line 56: I really like that the output-parsing is gone now 😬 !

src/formatting_stack/linters/eastwood.clj
```diff
@@ -12,14 +12,8 @@
...
-                        eastwood.lint/default-linters)]
-    (-> eastwood.lint/default-opts
-        (assoc :linters linters
-               :rethrow-exceptions? true))))
-
-(def parallelize-linters? (System/getProperty "formatting-stack.eastwood.parallelize-linters"))
+  (-> eastwood.lint/default-opts
+      (assoc :rethrow-exceptions? true)))
 
 (def config-filename "formatting_stack.clj")
```
vemv, line 18: (This file can be deleted)

src/formatting_stack/linters/eastwood.clj
```diff
@@ -53,8 +45,7 @@
                             :filename            (if (string? uri-or-file-name)
                                                    uri-or-file-name
                                                    (-> ^File uri-or-file-name .getCanonicalPath)))))
-         (concat (impl/warnings->reports output)
-                 (impl/exceptions->reports @exceptions)))))
+         (concat (impl/exceptions->reports @exceptions)))))
```
vemv, line 48: Redundant `concat`

vemv (commented):
Some notes to self, will fix along any feedback

.circleci/config.yml
thumbnail: I think this is a good change, but i'd rather implement it in terms of formatting-stack. I did some work to support cli invocation.

But i'd be fine with replacing this CI step when that's ready.

src/formatting_stack/linters/eastwood.clj
thumbnail: I don't think it is redundant, although it could be changed to `into`

vemv: Not sure why one would want to concat/into a coll alone?

.circleci/config.yml
vemv: I'd suggest having both for the hinted reason: using f-s to lint f-s is somewhat weak. It can self-report being OK while it's not.

One would run the self-linting before Eastwood though, for fast feedback.

This could be an entire sequence:

* f-s self-lint (kondo + eastwood, git diff only)
* clj-kondo, whole project
* eastwood, whole project

Maybe a little annoying for our standards of fast CI linting, but it could catch an issue the test suite didn't

thumbnail: In my suggestion it'd run the linter across the entire project as well. But we'll cross that bridge when we get there.

src/formatting_stack/linters/eastwood.clj
thumbnail: The concat is part of a thread-last operation, so it get's two arguments (reports from eastwood and exceptions).

I'd prefer `into` over `concat` because of it's eager nature.

vemv: whoopsie

.circleci/config.yml
vemv: There could be an issue in strategies, in reporting, in interaction with a given linter impl... bunch of things that can go wrong

vemv:
Amended to use `into`

vemv:
Removed the redundant config file. Strangely that revealed an unrelated failing test. I got to fix it by moving it to a different ns, prevening the git-cloning fixture from being run for it.

thumbnail:
> Strangely that revealed an unrelated failing test.

Yeah I looked into this for a bit and am not sure what is actually happening. Moving the test out of this test-ns works but doesn't fix the root cause (maybe the `refresh-within-refresh-dirs-only`-strategy doesn't work properly in the other git folder?). I'll look into this a bit further.

vemv:
The fixture was making `all-files (sut/all-files :files [])` return files such as these:

`/home/circleci/repo/git-integration-testing/src/formatting_stack/formatters/cljfmt.clj`

whereas they were expected to be like this:

`/home/circleci/repo/src/formatting_stack/formatters/cljfmt.clj`

...so it seems to me that `sut/namespaces-within-refresh-dirs-only` is doing its job properly as `git-integration-testing/src` is not a refresh dir.

thumbnail:
That makes sense; the test-fixture sets a different CWD to make sure the git-commands run in the right CWD; but that obviously doesn't work for the refresh dirs.

vemv:
Rebased and added a minor commit

vemv:
Added a commit following PR feedback, and removed the old one

src/formatting_stack/linters/eastwood.clj
```diff
@@ -12,36 +12,22 @@
...
-                     (binding [*warn-on-reflection* true]
-                       (try
-                         (cond-> options
-                           true                 (assoc :namespaces namespaces)
-                           parallelize-linters? (update :builtin-config-files conj config-filename)
-                           true                 (eastwood.lint/eastwood (impl/->TrackingReporter reports)))
-                         (catch Exception e
-                           (swap! exceptions conj e)))))]
+        exceptions (atom nil)]
+    (with-out-str
```
thumbnail, line 24: `formatting-stack.util/silence` might be helpful here, since you're not capturing the output:
	
https://github.com/nedap/formatting-stack/blob/6ca2d6ba6734bf6ecf1cdf5bb7b42adfa213b73d/src/formatting_stack/util.clj#L154

thumbnail (approved):
Seems good to me! Left 2 comments but nothing mayor.

src/formatting_stack/linters/eastwood.clj
```diff
@@ -12,36 +12,22 @@
    (java.io File)))
 
 (def default-eastwood-options
-  ;; Avoid false positives or undesired checks:
-  (let [linters (remove #{:suspicious-test :unused-ret-vals :constant-test :wrong-tag}
```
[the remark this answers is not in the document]
vemv: > Are all of these linters currently disabled by default in eastwood? What if they'll be re-enabled in the future?

They've been fixed upstream (i.e. they shouldn't give false positives) so there's no further need of disabling them.

(Which is why QAing a few large projects is recommended)

vemv:
Rebased again, now we have 3 commits. The first one includes the`silence` feedback

thumbnail:
Thanks a lot for the patience and the work!

Issue #204, opened 2022-03-03
Make `:formatting-stack/ns-aliases` understand a codebase's patterns


vemv:
## Context

While `:formatting-stack/ns-aliases` is useful, it also can be annoying to use in an existing codebase with no particular inclination towards using Sierra's style.

The ns style is sometimes outside one's control e.g. I'm checking out a random repo.

f-s could try to be smarter there and not complain so easily - it is part of the project's spirit to avoid giving a wall of inflexible warnings.

## Suggested task

A given project's `:acceptable-aliases-whitelist` is augmented with the set of aliases that are used in the project and not new (per git: they don't belong to the git status or the git branch diff)

## Acceptance criteria

* `:acceptable-aliases-whitelist` are augmented in a way that doesn't compromise overall accuracy
* the project-wide analysis is cached

thumbnail:
alternatively, with greater configurability we could disable it by default and allow consumers to turn it on in their repo.

vemv:
Sure thing, however what I had in mind is that one f-s can help one use Sierra's style for 'his' code, while not complaining at everything else.

It's pretty common in projects that there's no convention whatsoever for ns aliases (or that the convention is precisely 'do whatever you want').

So that's where a hybrid approach makes sense.

PR #205, opened 2022-03-08
Augment a given project's `:acceptable-aliases-whitelist` with the aliases it already uses

vemv:
## Brief

Fixes https://github.com/nedap/formatting-stack/issues/204. The tldr is:

> A given project's `:acceptable-aliases-whitelist` is augmented with the set of aliases that are used in the project and not new (per git: they don't belong to the git status or the git branch diff)

Needs refactor-nrepl latest ([see](https://github.com/clojure-emacs/refactor-nrepl/pull/372)) but degrades gracefully.

## QA plan

* Use f-s in a project that wasn't conforming to f-s' ns aliasing suggestions
* Observe that existing aliases, even if non-conforming, will be fine
* non-conforming new aliases will cause a linting fault.

This can be further verified by inspecting the `formatting-stack.defaults/default-linters` object, which will have a larger `:acceptable-aliases-whitelist` map.

## Author checklist

* [x] The build passes
  * [x] Correctness
  * [x] Test coverage
  * [x] Spec coverage
  * [x] Documentation
  * [x] Security
  * [x] Performance
  * [x] Breaking API changes

## Reviewer checklist

* [ ] I have reviewed the PR
  * [ ] Correctness
  * [ ] Test coverage
  * [ ] Spec coverage
  * [ ] Documentation
  * [ ] Security
  * [ ] Performance
  * [ ] Breaking API changes

### DIFF
--- a/src/formatting_stack/linters/ns_aliases/impl.clj
+++ b/src/formatting_stack/linters/ns_aliases/impl.clj
@@ -21,12 +21,8 @@
               m2))
 
 (def namespace-aliases-for*
-  "Not available if not using refactor-nrepl >= 3.4.1."
-  (when-let [f (try
-                 (requiring-resolve 'refactor-nrepl.ns.libspecs/namespace-aliases-for)
-                 (catch Exception _
-                   nil))]
-    @f))
+  (when (strategies/refactor-nrepl-3-4-1-available?)
+    @(requiring-resolve 'refactor-nrepl.ns.libspecs/namespace-aliases-for)))
 
 (defn namespace-aliases-for [files]
   (when namespace-aliases-for*
--- a/src/formatting_stack/linters/ns_aliases/impl.clj
+++ b/src/formatting_stack/linters/ns_aliases/impl.clj
@@ -1,0 +1,58 @@
+(ns formatting-stack.linters.ns-aliases.impl
+  (:require
+   [clojure.set :as set]
+   [clojure.spec.alpha :as spec]
+   [formatting-stack.strategies :as strategies]
+   [nedap.speced.def :as speced])
+  (:import
+   (java.io File)))
+
+(spec/def ::alias (spec/or :simple-alias symbol?
+                           :directive keyword?))
+
+(spec/def ::aliases (spec/coll-of ::alias :kind vector?))
+
+(spec/def ::project-aliases (spec/map-of symbol? ::aliases))
+
+(speced/defn ^::project-aliases merge-aliases [^::project-aliases m1, ^::project-aliases m2]
+  (merge-with (fn [x y]
+                (vec (reduce into #{} [x y])))
+              m1
+              m2))
+
+(def namespace-aliases-for*
+  "Not available if not using refactor-nrepl >= 3.4.1."
+  (when-let [f (try
+                 (requiring-resolve 'refactor-nrepl.ns.libspecs/namespace-aliases-for)
+                 (catch Exception _
+                   nil))]
+    @f))
+
+(defn namespace-aliases-for [files]
+  (when namespace-aliases-for*
+    (let [{:keys [clj cljs]} (namespace-aliases-for* files true)]
+      (merge-aliases clj cljs))))
+
+;; NOTE: this isn't necessarily a "strategy" (which would reside in the `strategies` ns),
+;; since it's composed of other strategy calls.
+;; This is more of a handy helper at the moment.
+(defn stable-files
+  "Files that already existed as of the default branch,
+  and that haven't been touched in the current branch"
+  []
+  (let [with (set (strategies/all-files :files []))
+        without (reduce into #{} [(strategies/git-diff-against-default-branch :files [])
+                                  (strategies/git-completely-staged :files [])
+                                  (strategies/git-not-completely-staged :files [])])
+        corpus (set/difference with without)]
+    (->> corpus
+         (mapv (speced/fn [^String s]
+                 (File. s))))))
+
+(def project-aliases (memoize
+                      (fn [{_cache-key :cache-key}] ;; there's a cache key for correct memoization
+
+                        ;; note that memoizing results is correct -
+                        ;; results don't have to be recomputed as the git status changes:
+                        ;; touching more files doesn't alter the fact that these aliases already were existing.
+                        (namespace-aliases-for (stable-files)))))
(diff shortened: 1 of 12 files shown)


test/functional/formatting_stack/linters/ns_aliases.clj
```diff
@@ -6,7 +6,8 @@
    [matcher-combinators.test :refer [match?]]))
 
 (deftest lint!
-  (let [linter (sut/new {:max-lines-per-ns 4})]
+  (let [linter (sut/new {:max-lines-per-ns 4
+                         :augment-acceptable-aliases-whitelist? false})]
```
vemv, line 10: Also cover the `true` path

src/formatting_stack/strategies.clj
```diff
@@ -80,11 +81,33 @@
        (impl/extract-clj-files)
        (into files)))
 
+(defn current-branch-name []
+  (-> (sh "git" "rev-parse" "--abbrev-ref" "HEAD")
+      (:out)
+      (string/split-lines)
+      (first)))
+
+(defn default-branch-name []
```
vemv, line 90: The introduction of this defn is kinda unrelated, but also necessary for this feature to work.

Anyway, the hardcoded `"master"` was very far from ideal, in fact f-s itself uses `main` as its default branch so it was pretty broken.

The new impl checks for `["master" "main" "stable" "dev"]`. One could also check `origin`, but IME git can opaquely hang over misc network issues. So having f-s hang would be a pain.

This defn could be improved in a more generalized direction in a future, but for now we can observe that it's better than before.

vemv:
Bumped refactor-nrepl again, which now has :as-alias support for clj 1.11.

src/formatting_stack/strategies.clj
thumbnail: I suppose it could also check a system property before making the assumption. If you also _set_ the system-property after it's cached and visible too

src/formatting_stack/linters/ns_aliases/impl.clj
```diff
@@ -0,0 +1,58 @@
...
+;; since it's composed of other strategy calls.
+;; This is more of a handy helper at the moment.
+(defn stable-files
+  "Files that already existed as of the default branch,
+  and that haven't been touched in the current branch"
+  []
+  (let [with (set (strategies/all-files :files []))
+        without (reduce into #{} [(strategies/git-diff-against-default-branch :files [])
+                                  (strategies/git-completely-staged :files [])
+                                  (strategies/git-not-completely-staged :files [])])
```
thumbnail, line 46: alternatively; `(into #{} cat [(strategies ...) (strategies ...)])`

thumbnail (commented):
Thanks for this one 🚀 


Some minor comments, addressable at your own discretion

src/formatting_stack/linters/ns_aliases/impl.clj
```diff
@@ -0,0 +1,58 @@
...
+(spec/def ::project-aliases (spec/map-of symbol? ::aliases))
+
+(speced/defn ^::project-aliases merge-aliases [^::project-aliases m1, ^::project-aliases m2]
+  (merge-with (fn [x y]
+                (vec (reduce into #{} [x y])))
+              m1
+              m2))
+
+(def namespace-aliases-for*
+  "Not available if not using refactor-nrepl >= 3.4.1."
```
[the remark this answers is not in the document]
vemv: The comment was correct, you tripped on the double negative ^^

The feature relies on a new API.

test/unit/formatting_stack/linters/ns_aliases/impl.clj
```diff
@@ -0,0 +1,19 @@
...
+
+(deftest merge-aliases
+  (are [m1 m2 expected] (match? expected
+                                (sut/merge-aliases m1 m2))
+    {}                      {}                      {}
+    {'a ['b]}               {}                      {'a ['b]}
+    {}                      {'a ['b]}               {'a ['b]}
+    {'a ['b]}               {'a ['c]}               {'a (matchers/in-any-order ['b 'c])}
+    {'a ['b]}               {'a ['b]}               {'a ['b]}
+    {'a ['b 'c] 'd ['e 'f]} {'a ['g 'h] 'd ['i 'j]} {'a (matchers/in-any-order ['b 'c 'g 'h])
```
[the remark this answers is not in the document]
vemv: I guess I don't like that one can turn `{'a` into `'{a` opportunistically.

An homogeneous pattern is simpler to compare visually/cognitively, else it might not be clear that you're comparing stuff of the same kind

src/formatting_stack/strategies.clj
vemv: IDK, instructing people to set a sysprop seems far from ideal.

In a future we can simply refine this defn. It might be possible to do the op offline (remote info is stored locally, except in CI)

vemv:
Cheers ☕️☕️

src/formatting_stack/strategies.clj
```diff
@@ -80,11 +81,28 @@
...
+  (let [all-branches (->> (sh "git" "branch")
+                          :out
+                          string/split-lines
+                          (map (fn [s]
+                                 (-> s (string/split #"\s+") last)))
+                          (set))]
+    (or (some all-branches ["master" "main" "stable" "dev"])
+        ;; return something, for not breaking code that traditionally assumed "master":
+        "master")))
```
thumbnail, line 99: By returning a known-bad value (master wasn't in `all-branches`, so it's not the right one) the diff-functionality doesn't work iirc.

I think highlighting that mistake / error is preferable to silently doing the wrong thing. With the system-prop suggestion above the error could even suggest setting the system-prop to solve the error.

This could be either an exception; or a simple `log/warn` in the `or`-case. WDYT?

src/formatting_stack/strategies.clj
vemv: > This way your addition would work exactly in the same way, but there's an escape hatch to configure this.

Alrighty!

src/formatting_stack/strategies.clj
vemv: We can also observe that no behavior is being changed. I wouldn't be super enthusiastic about breaking it on purpose.

The 'compromise' solution would be to return `"master"` _and_ log a warning (with the sysprop suggestion). sgty?

thumbnail: That sounds good 👌🏻

vemv:
Ready again!

thumbnail (approved):

